Issue #21, opened 2014-10-14
distributed-map locking examples fail


karlroberts:
The Non Racy examples fail.

re-create:- run PessimisticUpdateMember twice at the same time, neither result is the  expected 2000.

I have found the problem and will submit a pull request to the Maintenance branch.

amoAHCP:
Hi, you should really fix this. This "bug" is really confusing, I thought the issue is in the 3.8-SNAPSHOT code, so I replaced it by the stable 3.6.4 which also did not worked. I had a similar issue before... but I don't really get why @karlroberts solution solves the problem. THX

PR #23, opened 2014-10-14
Fix/3.x/21

karlroberts:
Fix for #21 

setup() tests whether the Map Value needs to be initialised before the test starts inside a lock block.

also shutdown hz Instance after the test, it is anoying that the demo leaves the Instance running which can conflict with other demos.


Holmistr:
@mmedenjak Any plans for this PR?

PR #37, opened 2015-02-10
hibernate-jpa-2ndlevel-cache. 2nd level cache enabled. JMX bean registra...

senleft:
The initial example contains only cache configuration without objects that used 2nd level cache.
In this pull request:
- @Caheable for Employee class added;
-  access to Hibernate statistics via jconsole added;
- 'show' command added, may used for see presence/absence SQL queries for not cached/cached entities.

### DIFF
--- a/hazelcast-integration/hibernate-jpa-2ndlevel-cache/README.md
+++ b/hazelcast-integration/hibernate-jpa-2ndlevel-cache/README.md
@@ -83,4 +83,9 @@ Key:
 1
 ```
 
+After application running it creates sample db with 19 records (id range from 1 to 19).
+If execute 'show' command with id in range above you will not see any SQL statements in the console (cached entities).
+
+4) Also you may see Hibernate statistics via jconsole.
+
 
--- a/hazelcast-integration/hibernate-jpa-2ndlevel-cache/pom.xml
+++ b/hazelcast-integration/hibernate-jpa-2ndlevel-cache/pom.xml
@@ -12,7 +12,7 @@
     <modelVersion>4.0.0</modelVersion>
 
     <artifactId>hibernate-jpa-2ndlevel-cache</artifactId>
-    <name>Hibernate 2nd Level Cache</name>
+    <name>Hibernate JPA 2nd Level Cache</name>
 
     <dependencies>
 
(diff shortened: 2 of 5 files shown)


bilalyasar:
@senleft , thanks for your contribution. Also we have HazelcastAccessor for hibernate, can you please get HazelcasInstance and show cache size somehow. (you can add new command for cache size or another thing...)

senleft:
@bilalyasar, I'm not sure if I understand you. I think that the hibernate-jpa-2ndlevel-cache sample goals are:
- explain how configure cache (actual example code do it).
- show that caching works (actual example code not do it, this pull request fix it).


Also I think it would be nice to show the statistics about Hibernate 2nd level cache, so in this pull request all statistics available via JMX or directly via org.hibernate.stat.Statistics API. If you will run ManageEmployeeJPA  it prints to console some of metrics, including the Hibernate 2nd level cashe size, after each command execution. The HazelcastInstance is not need to get it.

Another interesting thing discovered. If you remove  
```
<property name="javax.persistence.sharedCache.mode" value="ENABLE_SELECTIVE"/>
```
from persistence.xml then cashing still works, however
```
statistics.getSecondLevelCacheStatistics(regionName)
```
returns null.

I will appreciate you for explanation how
```
HazelcastInstance hazelcastInstance = HazelcastAccessor.getHazelcastInstance(sessionFactory);
```
helps to get all metrics that org.hibernate.stat.Statistics has.

bilalyasar:
@senleft what i mean is that second level caches are stored in hazelcast IMap, after getting instance you can get hazelcast Map.
Only map size, i think enough for me. Because some users may want to look cache map size. Also map is configurable.
example of getting map: https://github.com/hazelcast/hazelcast/blob/master/hazelcast-hibernate/hazelcast-hibernate3/src/test/java/com/hazelcast/hibernate/HibernateStatisticsTestSupport.java#L135

senleft:
@bilalyasar, Thank you very much for explanation. Please view the pull request second commit.

bilalyasar:
@senleft thanks for update..
@tugrulaltun has reviewed your pr. it is OK. and i will merge your PR but before the merging can you please sign this document:
https://hazelcast.atlassian.net/wiki/display/COM/Hazelcast+Contributor+Agreement
our all contributors have to sign these documents.

bilalyasar:
@senleft , thanks a lot :+1:

senleft:
@bilalyasar, You are welcome!

PR #38, opened 2015-02-12
Hazelcast cluster on Amazon EC2 with Vagrant and Chef

dbrimley:


bilalyasar:
i reviewed and tried this example.. it looks good. so i am merging it.

gAmUssA:
@dbrimley As a side note, in future, I would recommend to squash all commits into one before PR. So it would be easy in future to rollback in case of a problems. Also it will keep history of master branch clean.

mesutcelik:
:+1:

jerrinot:
+1 to @gAmUssA

PR #43, opened 2015-05-12
pom.xml fixed and readme.md updated

muratayan:
- As far as I tested it works as expected. I just changed the configuration for `pom.xml` so that it works directly when people download it. No need for extra work to add plug-in.

- Read me file updated. Added reminders for `licence keys` and some editorial changes.


bilalyasar:
:+1:

PR #45, opened 2015-05-12
Fixes in HD memory examples after code review

gAmUssA:
As per discussion and review with @serkan-ozal, made couple fixes in previously merged PR:
- removed HD near cache config
- used `EnterpriseSerializationService` to access native memory stats


serkan-ozal:
:+1:

PR #53, opened 2015-06-29
XML fixed so that cache returns same object

muratayan:
In order to get same object reference with second get operation, we need to specify `in memory format` as `OBJECT` and for caching to work in a single node we need to enable the `cache local entries` option.


emrahkocaman:
Murat, can you send this also to `maintenance-3.x` branch?

PR #58, opened 2015-09-04
Cleanup of clients module

Donnerbart:


Donnerbart:
There was and is an issue with the REST client example. It throws the following exception:
```
Value for key1 is value1

Exception in thread "main" java.io.StreamCorruptedException: invalid stream header: FFFFFFE9
	at java.io.ObjectInputStream.readStreamHeader(ObjectInputStream.java:806)
	at java.io.ObjectInputStream.<init>(ObjectInputStream.java:299)
	at RestClient.main(RestClient.java:41)
```
The same exception is thrown in the master branch, so it's not a problem of the refactoring. The problematic line is:
```
objectInputStream = new ObjectInputStream(byteArrayInputStream);
```

clients/client-rest/src/main/java/RestClient.java
```diff
@@ -0,0 +1,62 @@
...
+            objectInputStream = new ObjectInputStream(byteArrayInputStream);
+
+            Person person = (Person) objectInputStream.readObject();
+            System.out.println("Object for key1 is " + person.toString());
+        } finally {
+            closeQuietly(objectInputStream);
+            closeQuietly(byteArrayInputStream);
+        }
+
+        HazelcastClient.shutdownAll();
```
gAmUssA: @Donnerbart Why we need this here? I don't use any Hazelcast client functionality

Donnerbart: Sorry, copy&paste from the other clients test, where I used this instead of `System.exit(0)` to shutdown properly.

gAmUssA:
Apparently we have a regression in a REST client module since 3.5. REST client example works fine with Hazelcast 3.4.x
This issue describes the problem https://github.com/hazelcast/hazelcast/issues/6144

PR #69, opened 2015-10-26
Code sample for custom attributes & extractors

tombujok:


pveentjer:
Does this example make sense? Because you are extracting age, but extracting a regular field from an object was already provided; no need for a custom extractor.

tombujok:
It just shows how to use an extractor. IMHO it doesn't matter that I can extract this field via reflection since I'm extracting under different attribute's name. 
Can change it to ageInMonths and multiply the age by 12. wdyt?

pveentjer:
It is fine. I didn't see that the name of the field doesn't match the query. Can you add a line of comment to the example showing what is being tested so that people like me don't get confused?

tombujok:
Done. I changed it to `ageInMonths` and added a comment.

PR #70, opened 2015-10-26
Code sample for querying collections

tombujok:


distributed-map/query-collections/src/main/java/com/test/query/QueryCollectionsDemo.java
```diff
@@ -0,0 +1,54 @@
...
+import static com.test.query.QueryCollectionsDemo.Limb.limb;
+
+public class QueryCollectionsDemo {
+
+    public static void main(String[] args) {
+        HazelcastInstance hz = Hazelcast.newHazelcastInstance();
+        IMap<String, Person> map = hz.getMap("map");
+
+        map.put("1", new Person("peter", limb("left-hand"), limb("right-hand")));
+        map.put("2", new Person("hans", limb("linke-hand"), limb("rechte-hand")));
```
pveentjer: should this not be left-hand and right hand?

tombujok: Then both would match. I wanted just of persons to match.

pveentjer: The example is confusing to me because it looks like grammatical errors.

tombujok: I changed the german limb names to legs now, so both are in english now.

PR #72, opened 2015-11-04
OSGI code sample

serkan-ozal:


mesutcelik:
:+1:  had a quick look. seems ok. 

@cangencer ,
can you also have a look?

osgi/src/main/java/com/hazelcast/osgi/examples/SampleActivator.java
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * {@link org.osgi.framework.BundleActivator} implementation for OSGI sample which is activated by our sample bundle.
+ */
+public class SampleActivator implements BundleActivator {
+
+    @Override
+    public void start(BundleContext context) throws Exception {
+        final ClassLoader currentContextClassLoader = Thread.currentThread().getContextClassLoader();
+        try {
+            Thread.currentThread().setContextClassLoader(getClass().getClassLoader());
```
cangencer: I'm not quite clear why we need to swap the classLoader. Does the user of the service always need to swap classloaders before being able to create instances?

osgi/src/main/java/com/hazelcast/osgi/examples/SampleActivator.java
```diff
@@ -0,0 +1,69 @@
...
+
+            System.out.println("Starting activator " + this + " in bundle " + context.getBundle() + " ...");
+
+            // Find the service reference of `HazelcastOSGiService` instance
+            ServiceReference serviceRef =
+                    context.getServiceReference(HazelcastOSGiService.class.getName());
+            if (serviceRef == null) {
+                throw new IllegalStateException("There is no registered `HazelcastOSGiService`!");
+            }
```
cangencer: If possible using @Service annotation based injection instead would also be nice (maybe as a separate example?), as it's quite widespread.

osgi/src/main/java/com/hazelcast/osgi/examples/SampleActivator.java
serkan-ozal: Normally it is not (while I am testing it on both Eclipse and PAX (using Felix as embedded OSGI container in integration test), there is no problem) needed but in here there are two classloaders, application classloader and TCCL (bundle classloader), and both of them finds two entry for `ServiceLoader` and this leads to `ClassCastException`. I will look at it in detail.

serkan-ozal: `frameworkConfigs.put(Constants.FRAMEWORK_BUNDLE_PARENT, Constants.FRAMEWORK_BUNDLE_PARENT_APP);` fixed the hack

cangencer:
:+1:

PR #78, opened 2015-11-13
Reflect event filter changes

emrahkocaman:
Depends on https://github.com/hazelcast/hazelcast-enterprise/pull/469


dbrimley:
This code sample doesn't actually show you doing anything with the Entry.  It would be good to demonstrate some sort of decision being made here based on a property in the object.

emrahkocaman:
`SampleMapWanEventFilter` is updated to demonstrate using entry value.

PR #94, opened 2016-01-13
Spring Boot Hazelcast integration example.

gAmUssA:
Provided example of usage Spring Boot caching with Hazelcast via Spring Framework caching abstraction and JCache.


mustafaiman:
:+1:

PR #97, opened 2016-03-04
Code samples for JCache enhancements

serkan-ozal:
Includes:
* Code sample to demonstrate iterator usage on cache.
* Code sample to demonstrate near-cache behaviour when per entry invalidation is disabled.


emrahkocaman:
:+1:

PR #98, opened 2016-03-23
Fixed redirection in XSD schema URLs

Donnerbart:


jerrinot:
this does not look right. 
you can't change a target namesspace without changing the XSD itself. 
see for example the content of https://hazelcast.com/schema/config/hazelcast-config-3.6.xsd

and you cannot change a schema of a release version -> namespace URI has to stay.

PR #112, opened 2016-06-29
Nearcache Sample Fix

oguzdemir:
Waiting time for invalidation is increased to conform invalidation message batch frequency. Also, 5000ms margin is added to the waiting time.


serkan-ozal:
👍

PR #125, opened 2016-08-25
Added LicenseUtils to ease settings of Hazelcast Enterprise License key in EE code samples

Donnerbart:
* Added `LicenseUtils` to ease settings of Hazelcast Enterprise License key in EE code samples
* Fixed settings of license key in some code samples (was not set at all or forgotten for e.g. clients)
* Added JavaDoc to all EE code samples to advise how to set the EE license key

This replaces the four different methods to set a license key, which we had before:
* via `static final String` field in the code sample
* via local variable in the code sample
* via system property `hazelcast.enterprise.license.key`
* via XML config file `<license-key>YOUR_LICENSE_KEY</license-key>`

The new approach is a bit more hidden, but you only have to set the license key once to make it work for all Enterprise code samples. You can still override a `static final String` field in `LicenseUtils` and you can still use the system property. The new method is to put your license key into a file in your home directory, which I think is the most convenient way.

JavaDoc example:
```
/**
 * You have to set your Hazelcast Enterprise license key to make this code sample work.
 * Please have a look at {@link com.hazelcast.codesamples.helper.LicenseUtils} for details.
 */
```

### DIFF
--- a/enterprise/client-ssl/src/main/java/Client.java
+++ b/enterprise/client-ssl/src/main/java/Client.java
@@ -6,21 +6,25 @@ import com.hazelcast.core.HazelcastInstance;
 import java.io.File;
 import java.util.concurrent.BlockingQueue;
 
+import static com.hazelcast.codesamples.helper.LicenseUtils.ENTERPRISE_LICENSE_KEY;
+
+/**
+ * You have to set your Hazelcast Enterprise license key to make this code sample work.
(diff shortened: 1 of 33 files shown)


jerrinot:
looks could. could you add readme.txt / package-info.java into the helper module?

Donnerbart:
I created a `README.md`, like we have it most other modules (no `package-info.java` so far).

jerrinot:
:+1:

PR #131, opened 2016-10-03
Ring buffer store code sample

mmedenjak:
Code sample for ring buffer store

### DIFF
--- a/distributed-collections/pom.xml
+++ b/distributed-collections/pom.xml
@@ -29,6 +29,7 @@
         <module>itemlisteners</module>
         <module>list</module>
         <module>queuestore</module>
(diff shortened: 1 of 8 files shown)


jerrinot:
@mmedenjak: does the sample use a combination of declarative and programmatic configuration? that's not very clear from a user perspective.

mmedenjak:
I've removed the hazelcast.xml since I believe most examples are Java config based.

gAmUssA:
@mmedenjak xml config should be in examples as well. A lot of people use xml configs. Example should include both. Thank you

mmedenjak:
@gAmUssA Added.

PR #132, opened 2016-10-23
Add chemistry example and other modules to pom

neilstevenson:
Include Spark and Spring Data modules in buidl

### DIFF
--- a/hazelcast-integration/README.md
+++ b/hazelcast-integration/README.md
@@ -10,6 +10,8 @@
 	Implementation of spring cache provider integration of hazelcast as client-server application.
 - <h3>spring-configuration</h3>
 	Implementation of spring framework integration of hazelcast.
+- <h3>spring-data-hazelcast-chemistry-sample</h3>
+	Spring Data client-server example, based on the periodic table.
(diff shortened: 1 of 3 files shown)


gAmUssA:
@Donnerbart excluded those because they require Java 8

Donnerbart:
Yes, unfortunately we have the Java 6 requirement on all builds, so those modules weren't compiling. I tried to make them Java 6 compatible, but that didn't work out. Until we have an official go to change the Java version for the build job, we cannot activate them, otherwise the build job will crash with compile errors.

neilstevenson:
@gAmUssA @gAmUssA  Yeah, makes sense. A `hazelcast-code-samples-java8` would be the way forward, but probably more trouble than benefit.

Donnerbart:
It's "just" some integrations which actually need newer Java versions. All "internal" code samples should be easily compatible with Java 6, since Hazelcast itself is. So maybe we can just allow it for he integration module. If we limit the other modules to Java 6 via Maven, we still should see compile errors due to Java 7/8 usage in the modules, where we don't want to have this.

PR #144, opened 2017-01-09
Tidy up module definitions so they play nicely in IntelliJ.  Also put…

dbrimley:
… in order.

### DIFF
--- a/hazelcast-integration/pom.xml
+++ b/hazelcast-integration/pom.xml
@@ -46,9 +46,13 @@
         <module>jca-ra</module>
         <module>manager-based-session-replication</module>
         <module>mongodb</module>
+        <module>pcf-integration</module>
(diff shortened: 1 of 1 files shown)


Donnerbart:
These modules were excluded for a reason: They need Java > 6, which is NOT supported by our Jenkins build job at the moment. So the build will break and we'll get those nice notification emails again.

```
[ERROR] javac: invalid target release: 1.8
```

Will revert this until we agreed what to do with code samples and the Java version restriction.

jerrinot:
@Donnerbart oh, I didn't know that, sorry. 
Could we add the modules to a Maven profile which is activated by a newer Java version?

Donnerbart:
We discussed this after the stand-up and will allow Java 8 for the integration module. I will do the required changes today to get this working.

PR #149, opened 2017-02-16
Code samples for Scheduled Executor

tkountis:


distributed-executor/scheduled-executor-retrieve-all-tasks/src/main/java/BasicTask.java
```diff
@@ -0,0 +1,32 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import java.io.Serializable;
+import java.util.concurrent.Callable;
+
+public class BasicTask implements Callable, Serializable {
```
Donnerbart: Name suggestion: `EchoTask` (since it just returns its value)

Would be the same for all other `BasicTask` classes.

PR #150, opened 2017-03-02
Fix sample

snicoll:
This commit tries to clarify the sample with regards to some discussion
on the Spring Boot issue tracker, in particular:

https://github.com/spring-projects/spring-boot/issues/8275

To be able to reuse an `HazelcastInstance` in server mode, the location
of the configuration file should be set in both auto-configuration
unfortunately: there is currently now way to reuse an instance that was
loaded via the default configuration. See the following issue:

https://github.com/hazelcast/hazelcast/issues/10007

In a client scenario, the only way is to effectively disable the
auto-configuration which annoys me a lot. Probably Spring Boot should
back off automatically in such a scenario but we need a way to reliably
figure this out.

Also, I wasn't aware that Hazelcast provides different `CachingProvider`
that are looking for different things. IMO, there is a gap in the
caching support in Spring Boot about it and I'd be more than happy to
improve it as long as I don't need to resort to Hazelcast specific API
inside the JCache auto-configuration

### DIFF
--- a/hazelcast-integration/springboot-caching-jcache/src/main/java/com/hazelcast/springboot/caching/BootifulClient.java
+++ b/hazelcast-integration/springboot-caching-jcache/src/main/java/com/hazelcast/springboot/caching/BootifulClient.java
@@ -20,7 +20,6 @@ package com.hazelcast.springboot.caching;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.CommandLineRunner;
-import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.boot.autoconfigure.hazelcast.HazelcastAutoConfiguration;
 import org.springframework.boot.builder.SpringApplicationBuilder;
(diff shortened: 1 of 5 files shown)


snicoll:
anyone?

vbekiaris:
Hi @snicoll , apologies for missing this PR. Setting the instance name & explicitly setting the hazelcast XML configuration look good to me (in the context of JCache - `HazelcastInstance` resolution issues), however with this PR when I execute the member and then the client startup scripts, I see 2 Hazelcast members started.
/cc @neilstevenson

snicoll:
Oh well, looks like I missed something then. Let me take another look to it.

gAmUssA:
@snicoll did you have a chance to check @vbekiaris's comment?
/cc @neilstevenson

snicoll:
@gAmUssA I did and I forgot what the outcome was. Let me check again.

snicoll:
It indeed starts two clients and I can't remember what my intention was initially. I still think we should work on this sample because excluding the auto-config is wrong. Can you help with that?

snicoll:
I’ve asked for help and didn’t get any feedback. If you want to resume this we can have a look. It’s probably outdated but still relevant.

Holmistr:
Ok, thanks @snicoll . Sorry about that. @vbekiaris Could we resume this PR?

vbekiaris:
Sure, I'll get back on it in the next couple of weeks

vbekiaris:
@snicoll I have updated this code sample for 4.0 in #383. I think this PR is now obsolete, could you take a look?

Holmistr:
Given @vbekiaris 's comment, I'm closing this PR :) Thank you very much @snicoll !

PR #154, opened 2017-03-14
OCP integration demo and sample codes

lazerion:


lazerion:
need to add maven compiler properties

PR #162, opened 2017-04-20
hazelcast-integration/springboot-caching change to see cluster node s…

chrissekaran:
…cluster wide synchronization with cache locking


gAmUssA:
@chrissekaran could you explain reason for this PR?

chrissekaran:
It was meant to show someone the effect of using a lock. And the change needed to do so by the diff. 
So we don't need to merge it actually
Il close it

PR #164, opened 2017-04-21
Add NO_MIGRATION cluster state sample

mdogan:
See https://hazelcast.atlassian.net/wiki/display/PM/Gigantic+Cache+Migration+Enhancements requirement 3.


Holmistr:
@jerrinot The sample won't compile since the name of the public class is different from the filename. I will issue a fix for that right away, but we should be more careful next time to not introduce this type of failures IMHO :)

PR #177, opened 2017-08-03
[PREVIEW] Adding tests for samples

Holmistr:
Do not integrate this PR yet, this is a preview for a discussion.  I've discussed with @tombujok and @jerrinot , they agreed that this a good way to go, however, we need to make sure that it doesn't affect readability of the samples - hence this preview PR. Take a look and let me know what you think.


Holmistr:
Closing this based on the discussion :)

PR #178, opened 2017-08-08
Complete samples review, fixes + readme.html update

Holmistr:
I went through all the samples except integrations (therefore 28 samples skipped) and tried to run them. I found several issues, some of them minor, some of them causing the sample not working and I fixed them all. All the samples (except integrations) are working. 

I also updated the readme.html to better reflect what the user can find in the repository.


distributed-map/criteria-api/src/main/java/PredicateMember.java
```diff
@@ -9,10 +9,7 @@
 import java.util.HashSet;
 import java.util.Set;
 
-import static com.hazelcast.query.Predicates.and;
-import static com.hazelcast.query.Predicates.equal;
-import static com.hazelcast.query.Predicates.not;
-import static com.hazelcast.query.Predicates.or;
+import static com.hazelcast.query.Predicates.*;
```
Donnerbart: We don't use star imports in Hazelcast code, actually CheckStyle should fail on this.

screenshot from 2017-08-08 17-18-47
That's one of the few things you have to change in the IntelliJ code formatter to make it Hazelcast compatible.

Holmistr:
@Donnerbart I have fixed my Idea settings and fixed all the imports, thanks for noticing.

By the way, what checkstyle are you referring to? I don't see any Maven checkstyle plugin declarations, whole mvn clean install runs smoothly for me. I'm probably missing something.

Donnerbart:
CheckStyle is not part of the normal build profile. You can invoke it via (in most of our projects):
`mvn clean verify -Pcheckstyle`

I just see that at least the module `client-statistics` doesn't work, because there are properties missing in its `pom.xml`:
```xml
    <properties>
        <!-- needed for checkstyle/findbugs -->
        <main.basedir>${project.parent.parent.basedir}</main.basedir>
    </properties>
```
I can make a CheckStyle cleanup sweep after this PR has been merged.

PR #193, opened 2017-10-11
Adds dynamic datastructure configuration code sample

vbekiaris:


Donnerbart:
Did you check CeckStyle locally?

learning-basics/configure-datastructures-at-runtime/src/main/java/ClientDynamicConfig.java
```diff
@@ -0,0 +1,46 @@
...
+    public static void main(String[] args) {
+        // Start a member with no explicit configuration
+        HazelcastInstance instance = Hazelcast.newHazelcastInstance();
+        ILogger logger = instance.getLoggingService().getLogger(ClientDynamicConfig.class);
+
+        // Start a client -- we will be using the map to access & configure data structures
+        HazelcastInstance client = HazelcastClient.newHazelcastClient();
+        // Obtain a map
+        IMap<String, String> defaultMap = client.getMap("defaultMap");
+        // no data have been added yet, get("1") will return null
```
Donnerbart: `// no data has been added yet, so get("1") will return null`

vbekiaris:
Surprisingly, checkstyle passes even without any license header on the new files. The problem with `checkstyle` profile was wrong definition of `main.basedir` property.

learning-basics/configure-datastructures-at-runtime/src/main/java/ClientDynamicConfig.java
vbekiaris: You may be interested in the "is data singular or plural" debate 😉  : https://www.theguardian.com/news/datablog/2010/jul/16/data-plural-singular

PR #196, opened 2017-10-13
Improve distributed-map/backup sample

kwart:
Let the `distributable-map/backup` example show how the IMap behaves with backup and TTL configured.

The code newly uses JMX to show map statistics over a half of minute.


distributed-map/backup/src/main/java/Cluster.java
```diff
@@ -0,0 +1,54 @@
...
+
+        IMap<String, String> map = hz.getMap("cities");
+        map.put("where to live", "Brno");
+        map.put("where to have a joy", "Paris");
+        map.put("where to go next", "London");
+        System.out.println("Entries were added to 'cities' IMap, current size is: " + map.size());
+
+        // time-to-live is set on the 'cities' map, let's do some statistics over the time
+        for (int i = 0; i < 15; i++) {
+            Thread.sleep(2000L);
```
Donnerbart: You can use `com.hazelcast.examples.helper.CommonUtils.sleepSeconds` here, which makes it a bit easier to read (and maybe you also don't need the `throws Exception` then).

kwart: OK, thanks for looking into this. I'll fix it.

PR #200, opened 2017-10-17
Projectkey

neilstevenson:
Blog post on projections on Map Entry. key


neilstevenson:
It's building against Jet 0.5-SNAPSHOT, let's wait for 0.5 GA

PR #205, opened 2017-11-24
Fix Spring cahe-manager example - the Java package structure was wrong

kwart:
This PR fixes the Java packages in the `spring-cache-manager` sample. Instead of the single folder `com.hazelcast.spring.cache` the correct folder structure is used - `com/hazelcast/spring/cache`.


kwart:
Commit was extended. The same problem was also in the `spring-transaction-manager` sample.

PR #219, opened 2018-03-12
Fixes hazelcast/hazelcast#11872

emre-aydin:
Sample (`hazelcast-integration/spring-cache-manager/AnnotationBasedCacheManager`) 
fails to start because of a missing `PropertySourcesPlaceholderConfigurer`. 
One is registered to the context to fix the sample. 

Reproducer described in the issue is fixed when a
`Cacheable` bean and a `PropertySourcesPlaceholderConfigurer` is registered
to the context. This seems like a Spring Framework bug as when upgraded to a newer 
version (`4.2.9.RELEASE` for example) the problem goes away (although I didn't care
to find the relevant issue on Spring issue tracker).

Fixes: https://github.com/hazelcast/hazelcast/issues/11872

### DIFF
--- a/hazelcast-integration/spring-cache-manager/src/main/java/com/hazelcast/spring/cache/CachingConfiguration.java
+++ b/hazelcast-integration/spring-cache-manager/src/main/java/com/hazelcast/spring/cache/CachingConfiguration.java
@@ -3,12 +3,14 @@ package com.hazelcast.spring.cache;
 import com.hazelcast.client.HazelcastClient;
 import com.hazelcast.client.config.ClientConfig;
 import com.hazelcast.core.HazelcastInstance;
+
 import org.springframework.cache.CacheManager;
 import org.springframework.cache.annotation.CachingConfigurer;
 import org.springframework.cache.annotation.EnableCaching;
 import org.springframework.cache.interceptor.KeyGenerator;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.context.support.PropertySourcesPlaceholderConfigurer;
 
 @Configuration
 @EnableCaching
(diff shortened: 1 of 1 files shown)


emre-aydin:
Documentation says `Spring 2.5+` is supported, but the project depends on (and tests against) `4.3.0.RELEASE` -> https://github.com/hazelcast/hazelcast/blob/0a8833470e36465587e1ae3096808f4ee216367c/hazelcast-spring/pom.xml#L36

I'm not sure which version should be advertised as the minimum Spring version. We either need to test against older Spring versions as well or update the documentation to state we support `4.3.0.RELEASE+`.  I'm inclined to upgrade the version of this sample application so that it works (and revert addition of `PropertySourcesPlaceholderConfigurer`) and leave that discussion to elsewhere. There shouldn't be any reason not to use the latest 4.3.x version of Spring framework once upgraded to version 4 (BTW `4.2.0.RELEASE` onwards also work). I don't think it makes much sense to invest in fixing this for Spring 3.x as no one seems to have complained about it until now and it's a very old version of Spring. WDYT?

Small note: I don't think there's any problem specific to `AnnotationConfigApplicationContext.refresh()` method as the same problem occurs when you provide the `Configuration` classes to constructor of `AnnotationConfigApplicationContext` (which internally registers and refreshes the context).

gokhanoner:
@emre-aydin since docs says `Spring 2.5+ is supported`, this fix seems OK.

kwart:
@Serdaro What can we do with this on documentation level? Newer Spring versions have API changes (more methods to implement) older versions requires workarounds as this one for property replacement.

I can live with either fix, but it should be clear in documentation what and why should users implement to get the Spring caching using Hazelcast working.

mmedenjak:
@Serdaro @kwart is anything missing from this PR? Can we merge it or close it?

mmedenjak:
@emre-aydin can we merge this PR?

emre-aydin:
@mmedenjak yes we can merge it :+1:

PR #221, opened 2018-03-20
Added example for new merge policies

Donnerbart:
* ported `SplitBrainTestSupport` and `MergeBarrier` to the helper module
* ported `FireWallingConnectionManager` etc. classes to the helper module
* ported additional utility methods to the helper module
* added example for new merge policies

The `SplitBrainTestSupport` is not a 1:1 copy from the Hazelcast project, since we don't have the `TestHazelcastInstanceFactory` in the code samples. It's using the normal `HazelcastInstanceFactory` instead, but with a custom `FirewallingNodeContext`, to block and unblock the communication between nodes. It also doesn't provide a `createHazelcastInstanceInBrain(int brain)` method.

Nevertheless the code samples `SplitBrainTestSupport` is quite similar to the original, so it's very easy to convert existing split brain tests to code samples now.

Depends on https://github.com/hazelcast/hazelcast/pull/12573

### DIFF
--- a/helper/src/main/java/com/hazelcast/examples/helper/splitbrain/SplitBrainTestSupport.java
+++ b/helper/src/main/java/com/hazelcast/examples/helper/splitbrain/SplitBrainTestSupport.java
@@ -1,0 +1,402 @@
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.examples.helper.splitbrain;
+
+import com.hazelcast.config.Config;
+import com.hazelcast.core.Hazelcast;
+import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.core.HazelcastInstanceNotActiveException;
+import com.hazelcast.core.LifecycleEvent;
+import com.hazelcast.core.LifecycleListener;
+import com.hazelcast.examples.helper.nio.FirewallingConnectionManager;
+import com.hazelcast.examples.helper.nio.FirewallingNodeContext;
+import com.hazelcast.instance.HazelcastInstanceFactory;
+import com.hazelcast.instance.HazelcastInstanceImpl;
+import com.hazelcast.instance.HazelcastInstanceProxy;
+import com.hazelcast.instance.Node;
+import com.hazelcast.instance.NodeState;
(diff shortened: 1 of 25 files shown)


helper/src/main/java/com/hazelcast/examples/helper/splitbrain/SplitBrainTestSupport.java
```diff
@@ -0,0 +1,402 @@
...
+        onAfterSplitBrainCreated(brains.getFirstHalf(), brains.getSecondHalf());
+
+        healSplitBrain();
+        onAfterSplitBrainHealed(instances);
+    }
+
+    protected HazelcastInstance[] startInitialCluster(Config config, int clusterSize) {
+        HazelcastInstance[] hazelcastInstances = new HazelcastInstance[clusterSize];
+        for (int i = 0; i < clusterSize; i++) {
+            hazelcastInstances[i] = HazelcastInstanceFactory.newHazelcastInstance(config, createInstanceName(config),
```
Donnerbart: This creates a `HazelcastInstance` with a `FirewallingConnectionManager`, so we can block and unblock the communication between nodes.

mdogan:
Aren't code samples are just to show how a feature should/can be configured, how it should/can be used etc? I don't think they need to demonstrate the entirely working mechanism, such as creating real splits, failures etc.

Donnerbart:
It's an SPI feature, so we should show how to implement custom merge policies. And I think we need to demonstrate how they work. We already have this in the JCache module, but with a very unreliable method of creating a split-brain scenario.

mdogan:
I agree with showing how to implement a custom merge policy but to me, having a runnable demo with artificial split & merge goes beyond of code sample's main purpose. 
I think code samples should use only public API/SPI, from an end-user's perspective. They should not include & depend on any internal/private code from Hazelcast modules, which additionally brings a maintaining cost.

Donnerbart:
All code samples are executable, why should the merge policies be different? I think we _have_ to show how a custom merge policy really works (and if it works at all, I've already made some errors while implementing them). These code samples will also provide a nice playground to develop a custom merge policy and have a quick test run with it.

If we remove the JCache example, the maintenance may even be lower, since most of the code is now shared with our test infrastructure.

helper/src/main/java/com/hazelcast/examples/helper/splitbrain/SplitBrainTestSupport.java
```diff
@@ -0,0 +1,402 @@
...
+        return DEFAULT_BRAINS;
+    }
+
+    /**
+     * Override this method to create a custom Hazelcast configuration.
+     *
+     * @return the default Hazelcast configuration
+     */
+    protected Config config() {
+        return new Config()
```
mustafaiman: I believe these are for speeding up the tests. Not needed in a code sample.

helper/src/main/java/com/hazelcast/examples/helper/splitbrain/SplitBrainTestSupport.java
```diff
@@ -0,0 +1,402 @@
...
+
+    public void run() throws Exception {
+        try {
+            onBeforeSetup();
+
+            Config config = config();
+            int clusterSize = getClusterSize();
+            instances = startInitialCluster(config, clusterSize);
+
+            for (int i = 0; i < iterations(); i++) {
```
mustafaiman: I think we do not need iterations for the code samples.

PR #240, opened 2018-05-15
Update Spring Data version

leszko:


mesutcelik:
I think we have to move it to Spring Boot 2.0 when we release compatible version. 
see https://github.com/hazelcast/spring-data-hazelcast/pull/31

leszko:
👍 Yeah, when we realase 2.0, then it should also to moved to Spring Boot 2. Sure.

PR #242, opened 2018-05-24
Update Spring Data Code Samples to Spring Data Hazelcast 2.0

leszko:
This PR should be merged after the Spring Data Hazelcast 2.0.0 is released.


hazelcast-integration/spring-data-hazelcast-chemistry-sample/common/src/main/java/com/hazelcast/samples/spring/data/chemistry/CommonConfiguration.java
```diff
@@ -14,32 +10,4 @@
 @Configuration
 @EnableHazelcastRepositories(basePackages = {"com.hazelcast.samples.spring.data.chemistry"})
```
mesutcelik: I see the same annotation `@EnableHazelcastRepositories` in `AfterTranslatorConfiguration.java` too. Do we really need to define this two times?

leszko: You're right. Thanks Mesut!
I've removed the annotation from `AferTranslatorConfiguration.java`.

PR #252, opened 2018-08-03
Hazelcast for PCF with TLS enabled

googlielmo:
Sample Spring Boot Application to use Hazelcast for PCF with TLS enabled.


googlielmo:
verify

enterprise/pcf-tls-integration/README.md
```diff
@@ -0,0 +1,80 @@
+# pcf-tls-integration
```
leszko: Super minor: I think you don't need this line.

googlielmo: agree, a similar line was present in the pcf-integration README and I kept it... but yeah looks like it's just creating a confusing output

googlielmo:
Comments were addressed and tests passed. Merging.

PR #264, opened 2018-09-27
Remove unnecessary dependencies

jerrinot:
Deps are not used, they are probably left-overs from some
experiments.

### DIFF
--- a/clients/nodejs/package.json
+++ b/clients/nodejs/package.json
@@ -1,8 +1,6 @@
 {
   "dependencies": {
-    "hazelcast-client": "0.6.1",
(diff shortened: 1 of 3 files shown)


clients/nodejs/package.json
```diff
@@ -1,8 +1,6 @@
 {
   "dependencies": {
-    "hazelcast-client": "0.6.1",
-    "express": "^4.14.0",
-    "superagent": "^2.3.0"
+    "hazelcast-client": "0.6.1"
```
mustafaiman: We can update the client version to the latest `0.9.0`

PR #268, opened 2018-10-01
Adds samples for map.setTtl and cache.setExpiryPolicy

mustafaiman:


distributed-map/eviction/src/main/java/OverrideTtl.java
```diff
@@ -0,0 +1,70 @@
...
+
+    public static void main(String[] args) throws InterruptedException {
+        Config config = new Config();
+        config.getMapConfig("mapWithTtl").setTimeToLiveSeconds(1);
+
+        HazelcastInstance instance = Hazelcast.newHazelcastInstance(config);
+
+        // Entries in this map live until they are explicitly removed
+        IMap<Integer, String> defaultMap = instance.getMap("map");
+        // Entries in this map are removed after two seconds
```
vbekiaris: "two" -> "one" (or change the ttl in config above to 2 😉 )

PR #276, opened 2018-11-08
Update Spring Data Code Samples Spring Data version 2.1

leszko:
Merge after `spring-data-hazelcast` version `2.1` is released.


leszko:
verify

PR #277, opened 2018-11-08
Fix not building jib-maven-plugin

leszko:
Fixes the following issue which sometimes occurs while building the Docker image:
```
[ERROR] Failed to execute goal com.google.cloud.tools:jib-maven-plugin:0.9.13:build (default) on project jib-demo-project: Build image failed, perhaps you should use a registry that supports HTTPS or set the configuration parameter 'allowInsecureRegistries': Only secure connections are allowed, but tried to reach URL
```

### DIFF
--- a/hazelcast-integration/aws-ecs/pom.xml
+++ b/hazelcast-integration/aws-ecs/pom.xml
@@ -74,6 +74,9 @@
                 <groupId>com.google.cloud.tools</groupId>
                 <artifactId>jib-maven-plugin</artifactId>
                 <version>0.9.13</version>
+                <configuration>
+                    <allowInsecureRegistries>true</allowInsecureRegistries>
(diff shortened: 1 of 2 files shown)


mesutcelik:
What problem made you send the PR? It is good to mention it in Description.

leszko:
I've added the exception you may sometimes get while building the image. To be frank, when I tried it now, I couldn't reproduce it. If it were the production code, I'd probably dig deeper to not allow some security holes. However, since it's a code sample, and this part is just about building the image, I'd just merge. Otherwise, we may risk that some users try to walk through the sample, but can't build it.

@mesutcelik , wdyt?

mesutcelik:
I just wanted to see the motivation of the change as future reference. Thx for adding!

PR #285, opened 2018-12-18
added code sample for Client name and attributes usages

sancar:


clients/client-attributes/src/main/java/AttributesUsage.java
```diff
@@ -0,0 +1,46 @@
...
+        ClientConfig clientConfig = new ClientConfig();
+        //setting instance name and attributes
+        clientConfig.setInstanceName("SampleClientName");
+        clientConfig.setAttribute("role", "admin");
+        clientConfig.setAttribute("region", "foo");
+
+        HazelcastClient.newHazelcastClient(clientConfig);
+
+
+        //client opened with xml config. See ./resources/hazelcast-client.xml
```
ihsandemir: you can separate the xml and programatic version into two examples.

sancar: Done

sancar:
run-lab-run

PR #294, opened 2019-02-05
add code sample for fail-over/blue-green support of clients

sancar:


enterprise/client-failover-cluster/src/main/java/sample/DeclarativeFailoverExample.java
```diff
@@ -0,0 +1,42 @@
...
+import com.hazelcast.core.LifecycleListener;
+
+public class DeclarativeFailoverExample {
+
+    public static void main(String[] args) {
+
+        HazelcastInstance client = HazelcastClient.newHazelcastFailoverClient(null);
+
+        //Client will by default logs as info
+        //INFO: hz.client_0 [cluster1] [3.12-SNAPSHOT] HazelcastClient 3.12-SNAPSHOT (20190205 - 3af10e3) is CLIENT_CHANGED_CLUSTER
```
ihsandemir: 3.12-SNAPSHOT ???

sancar: It is an example log. Whatever, I put here will be out-of-date. 
What do you think I should write? What do you mean by `3.12-SNAPSHOT ???` ?

ihsandemir: I see that you already removed the SNAPSHOT, which is good. I would prefer also the first sentence as: `Client will by default logs as info. The log will be similar to:`

PR #307, opened 2019-02-28
Kubernetes WAN replication

leszko:


leszko:
> I have gone through that. It works fine. What actually thought before starting the review is that this sample would use Kubernetes discovery SPI and smart client... Have you tested that or you assume it will work. If you tested already then good to refer to external client code sample or lets add the same content here.

Yes, it works fine with the External Client configuration. You can check the related [TDD ](https://hazelcast.atlassian.net/wiki/spaces/PM/pages/1313933488/OpenShift-Kubernetes+WAN+Verification+Technical+Design+Document)for details. I added a note here.

PR #309, opened 2019-03-04
Add YAML configuration code samples

blazember:


learning-basics/configure-yaml/README.md
```diff
@@ -0,0 +1,43 @@
...
+- `ClasspathYamlConfig`: loads the configuration from the classpath based on the provided filename
+- `FileSystemYamlConfig`: loads the configuration from the file system based on the provided filename
+- `UrlYamlConfig`: loads the configuration from the provided URL
+- `InMemoryYamlConfig`: loads the configuration from the provided `String` containing YAML configuration 
+
+In the case of Java clients `ClientClasspathYamlConfig` can be used for the same purpose.
+  
+The sample `YamlConfigClasspath` loads the member configuration `hazelcast-sample.yaml` from the classpath and 
+demonstrates variable replacement in YAML configuration by defining the member's port in variable taken from 
+the system properties and then from a provided `Properties` instance.
```
leszko: Would it make sense to add the information on how to run this sample and what the reader should observe?

learning-basics/configure-yaml/src/main/java/ConfigLookup.java
```diff
@@ -0,0 +1,32 @@
...
+import com.hazelcast.core.HazelcastInstance;
+
+import static com.hazelcast.examples.helper.CommonUtils.sleepMillis;
+
+public class ConfigLookup {
+    public static void main(String[] args) {
+        // XML takes precedence: loading member configuration hazelcast.xml from the classpath
+        // despite the presence of hazelcast.yaml
+        HazelcastInstance member = Hazelcast.newHazelcastInstance();
+        sleepMillis(1000);
```
leszko: Why do you need this sleeps? For the user to observe the results? Or this is necessary for some reason? Could you maybe comment it in the code?

learning-basics/configure-yaml/src/main/resources/hazelcast-client.yaml
```diff
@@ -0,0 +1,6 @@
+hazelcast-client:
+  import:
```
leszko: I guess this import is for the reader to observe something, would it make sense to describe it in README?

learning-basics/configure-yaml/src/main/resources/hazelcast-client.yaml
```diff
@@ -0,0 +1,6 @@
...
+    - hazelcast-client-network-config.yaml
+  group:
+    name: dev-configured-from-yaml
```
leszko: Do you use it somehow in the code sample?

learning-basics/configure-yaml/README.md
blazember: The samples are implemented in main methods without using shell scripts. I can add that, but I think it's not a question for a java engineer.

What the user should observe when running the samples is added as inline comments.

learning-basics/configure-yaml/src/main/java/ConfigLookup.java
blazember: I used the `configuration-xml` sample as a basis, the sleeps were there. Removed, it wasn't necessary.

learning-basics/configure-yaml/src/main/resources/hazelcast-client.yaml
blazember: Added to the README that the first test demonstrates composing YAML configuration files.

learning-basics/configure-yaml/src/main/resources/hazelcast-client.yaml
blazember: It is needed to be the same in member and client config. It's better to be unique in the samples so no members and clients join from or to instances started by other samples by mistake.

blazember:
Thanks for the comments @leszko, addressed them. Could you please take a second look?

blazember:
Thanks for the review @lazerion @leszko :+1: 

Squashed and added a new section referencing the `hazelcast-full-example.yaml` configuration to the README.

PR #312, opened 2019-03-14
Fix Management Center namespace in OpenShift Code Sample

leszko:
Before it only worked if your project was named "hazelcast".


leszko:
Yeah, I agree, we could have one file and change the image with the parameters. Anyway, let's not do it right now.

PR #198, opened 2017-10-17
Simplify mapstore example - remove the JDBC part completely

kwart:
The mapstore sample combines more samples into a one. This PR removes the complex JDBC+Hibernate part and leaves just the simple `ConcurentHashMap`-based one.

The main issues in the JDBC part of the sample:
* neither a readme nor a run script is provided
* leaking JDBC Statements (closing missing)
* unable to replace value in the person map (due to constraints violation)

The Hibernate part is not used at all.

We should only have samples which are copy/paste ready quickstarts.

PR #315, opened 2019-03-15
Remove Spring Data Integration Code Sample.

leszko:
This Code Sample is related to MongoDB, not Hazelcast. Therefore, I does not make much sense to keep it in our repo. What's more, it's super misleading, because one may think it's related to our mode: spring-data-hazelcast.


mesutcelik:
That was confusing for me too all the time but removing might not be solution. It is an example of how MapStore/MapLoader is implemented. Maybe renaming could be an option. cc: @jerrinot @mmedenjak

leszko:
@jerrinot @mmedenjak , wdyt?

mmedenjak:
I must say I would like to keep _some_ sample of implementing a map store. Do we have some other sample?

mesutcelik:
There is one under `distributed-map`
https://github.com/hazelcast/hazelcast-code-samples/tree/master/distributed-map/mapstore

Which is waiting another simplification PR
https://github.com/hazelcast/hazelcast-code-samples/pull/198

I think we can move MongoDB and JPAStore into the same place...

leszko:
@mmedenjak , how about keeping only the simplified by @kwart : #198 ? I guess having just one (and the simplest possible) mapstore example is what we need.

mmedenjak:
Sure, if there's another (simplified) example, let's keep that one instead.

PR #318, opened 2019-04-04
Add Kubernetes Sidecar Code Sample

leszko:
A Code Sample which presents how Hazelcast can be used a Sidecar in Kubernetes.


mesutcelik:
Hi @leszko ,

I think that worked fine for me but the only issue was to push docker image to Docker Hub. I was sitting in a slow connection so I skipped that part and used minikube.
I think you can completely change it to minikube based sample or include minikube part too. I leave that to you... Here is what I have done to run it in minikube...

1. minikube start (needs to be installed if it is not already installed)
2.  `eval $(minikube docker-env)` before building docker image.. I also changed it to hazelcast/hazelcast-python-client not to offer everyone to change it to their own name.
3. `imagePullPolicy: IfNotPresent` added to deployment.yaml in order to instruct minikube not to try to pull the image but use it from local registery
4. `minikube ip` to find out local minikube ip
5. `kubectl describe svc hazelcast-sidecar` to find the nodeport
6. convert curl commands to `minikube_ip:nodeport` from item 4 and 5 above

Other than those, I think this will be incredibly useful code sample. Many Thanks!

leszko:
@mesutcelik I updated the Code Sample to include the informaion about minikube. PTAL.

I don't want to make it minikube-specific, but I gave some guidelines for users using Minikube and Docker Desktop.

hazelcast-integration/kubernetes/samples/sidecar/config.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: hazelcast-configuration
+data:
+  hazelcast.xml: |-
```
mesutcelik: What about changing that to yaml after upgrading 3.12?

leszko: Good idea, changed!

PR #322, opened 2019-05-23
bump master version to 4.0-SNAPSHOT

hasancelik:


kwart:
Merging as we see false negatives in the daily jobs.

PR #323, opened 2019-05-29
Fixed compilation issues after bumping version to hazelcast 4.0

tezc:


tezc:
Thanks for the review.

PR #328, opened 2019-06-17
Clean-up repositories in pom.xml to allow Maven artifact proxy utilization

kwart:


kwart:
Thanks for the review, @lazerion .

PR #334, opened 2019-07-10
Conforms breaking changes in 4.0

mustafaiman:


mustafaiman:
Thanks for the review @lazerion

PR #337, opened 2019-07-29
Reformat Hibernate code sample.

enozcan:
Simplified the code sample and corrected the second level cache logic.

The older one was not checking the status of secondary level cache but instead validating the database functionality only.


enozcan:
As I noticed a few glitches, let me make another pass on it before merge along with conflict changes.

pivovarit:
Enes, I think it's safe to merge it already. We should, of course, go for another iteration but since we have limited time, I believe that:

 _better done than perfect_

hazelcast-integration/hibernate-2ndlevel-cache/README.md
```diff
@@ -4,60 +4,81 @@ In this repository, you can find a sample implementation of hibernate 2nd level
 
 ## Prerequisites
 
-You should have installed Apache Maven(http://maven.apache.org/download.cgi).
+You should have installed Apache Maven and Mysql database.
```
pivovarit, line 7: > and Mysql database

We could probably simplify this step by preparing a set of Docker commands. Let's keep that in mind for the next iteration

enozcan: That's exactly what I mean by `glitches`. h2 db, for instance, would simply serve the purpose here without dealing any Mysql stuff. I will take care of them in the next iteration.

enozcan:
@pivovarit Is that code sample supposed to be 4.x compatible?

pivovarit:
Yes, we don't want to promote 3.x anymore

enozcan:
The PR is relatively old and we better proceed with #427.
Hence closing this.

PR #346, opened 2019-09-20
Renamed MapAttributeConfig and it's "extractor" field (#15547)

devozerov:


devozerov:
The plain countdown latch sample is removed completely because it referred to no longer existing API. The CP-based sample should be used instead.

devozerov:
run-lab-run

PR #351, opened 2019-10-08
Added a Hot Restart Encryption example

vojtechtoman:
Also cleaned up the generics etc. in the remaining
Hot Restart examples.


vojtechtoman:
run-lab-run

PR #353, opened 2019-10-15
Rename StartServer => HazelcastMemberStarter

leszko:
@Holmistr , please merge while relesing 4.0.


leszko:
> You can also update
> 
> https://github.com/hazelcast/hazelcast-code-samples/blob/8b09d10f3ff71a2c0cc465b6b62ded61376cdc14/network-configuration/jclouds/README.md#L19

@mmedenjak Good point! Updated.

Holmistr:
Given that the code samples are relying on 4.0-SNAPSHOT, I guess we should merge this already @leszko  WDYT?

leszko:
> Given that the code samples are relying on 4.0-SNAPSHOT, I guess we should merge this already @leszko WDYT?

Yes :)

PR #355, opened 2019-10-18
Fix 4.0 compilation issues

tezc:
I tried to fix compile but I'm not sure if changes are okay. Please take a close look to diff.
Changes are mostly in Security, RemoteService, Integration examples.

I'm open to suggestions, please either comment here if you want correction on some part or one can create another PR. Consider this PR as a coordination point :)

Note: 
Even compilation succeeds, some test cases are actually failing when you run them (e.g test cases are still using XML config for 3.x).  I believe it will be handled with a better approach to code samples later.


sancar:
Connection Retry is OK.

I think we can delete CounterService. SPI made hidden for now. In the Migration Guide we mentioned it. So I think, code sample should also be deleted. https://github.com/hazelcast/hazelcast-reference-manual/commit/b66ee13f9a2ca956dfd6619441cdee66ab24f5f1#diff-d49de7f68e55091944a162581ebc215dR63

tezc:
> Connection Retry is OK.
> 
> I think we can delete CounterService. SPI made hidden for now. In the Migration Guide we mentioned it. So I think, code sample should also be deleted. [hazelcast/hazelcast-reference-manual@b66ee13#diff-d49de7f68e55091944a162581ebc215dR63](https://github.com/hazelcast/hazelcast-reference-manual/commit/b66ee13f9a2ca956dfd6619441cdee66ab24f5f1#diff-d49de7f68e55091944a162581ebc215dR63)

Removed CounterService examples.

enterprise/pcf-tls-integration/src/main/java/com/hazelcast/pcf/integration/Application.java
```diff
@@ -64,12 +64,8 @@ public static void main(String[] args) throws IOException {
         List hazelcast = (List) json.get("hazelcast");
         Map map = (Map) hazelcast.get(0);
         Map credentials = (Map) map.get("credentials");
-        String clusterName = (String) credentials.get("cluster_name");
-        String clusterPassword = (String) credentials.get("cluster_pass");
         List<String> members = (List<String>) credentials.get("members");
-
         clientConfig = new ClientConfig();
-        clientConfig.setClusterName(clusterName).setClusterPassword(clusterPassword);
```
hasancelik: @kwart, me and @tezc reviewed these lines but we could not decide how we should configure it?
You have removed cluster-name and cluster-pass from hazelcast client code base:
https://github.com/hazelcast/hazelcast/commit/d2967c0e5f51d2f2cc91d80e71434b01c133632b#diff-193c80d2f8628d78a1ee54b494b1b869

`Application.java` starts client with parameters from PCF service that are defined by member side via json file:
https://github.com/hazelcast/hazelcast-code-samples/blob/master/hazelcast-integration/pcf-integration/hazelcast.json#L10-L13

I think we should remove password word from `Application.java` and above json file then configure client:
```
String clusterName = (String) credentials.get("cluster_name");
List<String> members = (List<String>) credentials.get("members");
clientConfig = new ClientConfig();
clientConfig. setClientName(clusterName);
```
Is this correct?

kwart: Probably correct.
I don't know how the JSON magic works here.

In XML we have - instead of
```
<group>
    <name>dev</name>
    <password>dev-pass</password>
</group>
```
there is the following in 4.0:
```
<cluster-name>dev</cluster-name>
```

There was also an intermediate state from @petrpleshachkov:
```
<cluster>
    <name>dev</name>
    <password>dev-pass</password>
</cluster>
```

hasancelik: Thanks for the info @kwart 👍 
@tezc, could you update the PR? I will fix json part later.

tezc: Apparently it is now setClusterName() : https://github.com/hazelcast/hazelcast/commit/572c7485fbe9560c9032f75ee7730c8ef9826184

I've updated the PR.

tezc:
Thanks for reviews, merging PR to fix compile.

PR #360, opened 2019-11-08
Adapt JCache code samples to changes in Hazelcast CachingProviders

vbekiaris:
- Previously public static factories have been removed and specific
Hazelcast server/client caching provider classes are private API.
Code samples are adapted to use public API where possible. In some
cases (eg split-brain code samples) it is impossible to use public
API; usage of private API methods is explicitly indicated in code
samples' javadoc
-  Default caching provider returned by Caching.getCachingProvider
is client-side caching provider. In order to obtain a server-side
caching provider, user has to explicitly request for that by
class name.


Holmistr:
@vbekiaris Checkstyle error :)

vbekiaris:
of course :)

vbekiaris:
thanks @Holmistr @tezc for reviews!

PR #364, opened 2019-11-13
Fix compilation after changes in client-side caching provider

vbekiaris:


vbekiaris:
thanks for the review @tezc !

PR #372, opened 2019-12-10
Multiple fixes for 4.0

tezc:


distributed-map/entry-processor/src/main/resources/hazelcast.xml
```diff
@@ -1,10 +1,16 @@
...
-                               http://www.hazelcast.com/schema/config/hazelcast-config-3.11.xsd"
+                               http://www.hazelcast.com/schema/config/hazelcast-config-4.0.xsd"
            xmlns="http://www.hazelcast.com/schema/config">
     <network>
         <join>
             <multicast enabled="true"/>
         </join>
     </network>
+
+    <user-code-deployment enabled="true">
```
mmedenjak: How come this is enabled?

tezc: I don't even remember why I needed it. Removed in another commit.

PR #376, opened 2019-12-16
Fixes of JCache integration + removing client artifacts reference

Holmistr:


Holmistr:
Need to merge it with red PR builder, since there's no deploy job for Beta release branch.

PR #389, opened 2020-02-18
Add basic EE WAN configuration sample

blazember:
The change makes a maven module level change by making `wan-replication`
to be a pom project under which the concrete WAN replication samples are
contained as separate maven modules. The original `wan-replication` sample
is moved under the `interactive-sample` module.

The new module `basic-sample` is introduced to present the most basic WAN EE
configuration possible in XML, YAML and programmatic ways. See the project's 
README for more details.


enterprise/wan-replication/basic-sample/src/main/resources/hazelcast-wan-source.xml
```diff
@@ -0,0 +1,66 @@
...
+    <license-key>YOUR_LICENSE_KEY</license-key>
+
+    <map name="wan-replicated-map">
+        <!--
+          The reference to the `wan-replication` configuration to be used for
+          replicating this map. The name should match to the name set in the
+          referenced `wan-replication` configuration element.
+         -->
+        <wan-replication-ref name="wan-sample">
+            <merge-policy-class-name>PassThroughMergePolicy</merge-policy-class-name>
```
mmedenjak: The merge policy should be optional (and I don't see it in the programmatic config) so maybe you can remove it from here and other occurrences in XML and YAML.

blazember: Yeah, I tried to remove the merge policy, but it resulted in an invalid YAML config, so I added it back. For now, I'd add it to programmatic too to have really the same config in the three ways. After fixing the YAML structure, I remove this.

blazember: Solved with this instead:
```
...
wan-replication-ref:
      wan-sample: {}
```
So the merge policy configurations could be removed. Also created https://github.com/hazelcast/hazelcast/issues/16671 to make it more straightforward to configure wan replication reference with defaults.

blazember:
Thanks for the review. I added the mentioning of active-passive mode in the readme and in the main classes. Could you please check if it made better for you?

blazember:
Thanks for the review @mmedenjak :+1:

Serdaro:
Thanks a lot for this @blazember

PR #395, opened 2020-03-05
Cleanup WAN code samples

mmedenjak:
- remove remnants of 3.x
- minor warning and code style fixes

### DIFF
--- a/enterprise/wan-replication/interactive-sample/src/main/resources/hazelcast.xml
+++ b/enterprise/wan-replication/interactive-sample/src/main/resources/hazelcast.xml
@@ -21,6 +21,21 @@
             <cluster-name>clusterB</cluster-name>
             <target-endpoints>127.0.0.1:5702</target-endpoints>
         </batch-publisher>
+        <!--
+        <custom-publisher>
+            <class-name>YOUR_PUBLISHER_CLASS_NAME</class-name>
+            <publisher-id>YOUR_PUBLISHER_ID</publisher-id>
+            <properties>
+                <property name="key">value</property>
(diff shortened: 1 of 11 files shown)


enterprise/wan-replication/interactive-sample/src/main/resources/hazelcast.xml
```diff
@@ -21,24 +21,11 @@
...
-                <property name="key">value</property>
-            </properties>
-        </custom-publisher>
-        <consumer>
-            <class-name>YOUR_CONSUMER_CLASS_NAME</class-name>
-            <properties>
-                <property name="key">value</property>
-            </properties>
-        </consumer>
-        -->
```
blazember: The readme of this sample references these configs. We should either remove that as well or keep this. I think it's good that we mention this in the sample, so I'd keep it.

mmedenjak: Hm, ok, I'll bring them back although I'd prefer if we had a separate sample for custom replication, this way it was a bit puzzling to me.

blazember: I agree that having a separate custom sample is better. My point is to keep the config consistent with the readme. I'm happy to add a custom replication sample next week and cleaning this up.

mmedenjak: Reverted in a separate commit.

mmedenjak:
Thanks for the review!

PR #396, opened 2020-03-06
Update distributed collections code samples

mmedenjak:
- cleanup warnings in code
- bump XMLs to 3.12 for examples for which I am not unsure if they can
be upgraded to 4.0

### DIFF
--- a/jcache-1.1/times-table/pom.xml
+++ b/jcache-1.1/times-table/pom.xml
@@ -24,8 +24,8 @@
         <hazelcast.version>${profile.hazelcast.version}</hazelcast.version>
         <javax-cache.version>${profile.jsr107.version}</javax-cache.version>
 
-        <java.version>8</java.version>
-        <spring-boot.version>1.5.9.RELEASE</spring-boot.version>
+        <java.version>1.8</java.version>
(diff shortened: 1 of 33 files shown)


jcache-1.1/times-table/pom.xml
```diff
@@ -137,7 +137,7 @@
                 <activeByDefault>true</activeByDefault>
             </activation>
             <properties>
-                <profile.hazelcast.version>3.9.3</profile.hazelcast.version>
+                <profile.hazelcast.version>3.12.6</profile.hazelcast.version>
```
mmedenjak: Because we have a profile using JSR 1.0.0 and HZ 3.9, I'm not sure that we can upgrade the code and XML to 4.0. On the other hand, maybe we can freely remove the code sample for JSR 1.0.0. WDYT @vbekiaris ?

vbekiaris: This code sample demonstrates a subtle difference in how you obtain caches in JCache 1.1 vs 1.0 (associated blog post https://hazelcast.com/blog/times-table-jcache-1-1/), so I wouldn't remove the `jcache10` profile (yet). I think keeping both profiles running on 3.y is best for as long as we keep this code sample.

PR #397, opened 2020-03-06
Update distributed map code samples

mmedenjak:
- cleanup warnings in code


Holmistr:
run-lab-run

PR #400, opened 2020-03-06
Update configuration code samples

mmedenjak:
- cleanup warnings in code
- bump XMLs to 3.12 for examples for which I am not unsure if they can
be upgraded to 4.0


network-configuration/jclouds-partitiongroup/pom.xml
```diff
@@ -23,6 +23,7 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
         <jclouds.version>1.9.1</jclouds.version>
+        <hazelcast.version>3.12.6</hazelcast.version>
```
mmedenjak: I think we have to keep jclouds examples at 3.12 because we haven't yet released the plugin compatible with 4.0 - https://github.com/hazelcast/hazelcast-jclouds/releases. Needs to be double-checked though.

network-configuration/ssl/pom.xml
```diff
@@ -11,15 +11,40 @@
     </description>
 
     <parent>
-        <groupId>com.hazelcast.samples.network-configuration</groupId>
-        <artifactId>network-configuration</artifactId>
+        <groupId>com.hazelcast.samples</groupId>
```
mmedenjak: SSL and symmetric encryption code samples require EE so it sucks that the code samples were using OS and failing to start.

network-configuration/jclouds-partitiongroup/pom.xml
Holmistr: Yes, it's like that. We can't upgrade that yet.

network-configuration/ssl/pom.xml
Holmistr: Shouldn't we rather move it to the `enterprise` directory? Just thinking

mmedenjak: Yes, you're right, I'll move it.

mmedenjak: Moved :)

kwart:
The moved modules will probably need an update in POMs to fix checkstyle config paths.

mmedenjak:
Fixed and addressed all comments @kwart @Holmistr

PR #407, opened 2020-03-11
Import kerberos-login-module

zpqrtbnk:
This PR imports sample code for a Kerberos login module, corresponding to Kerberos support in the .NET Client 4.0. This is a DRAFT PR, while it is being reviewed.

Notes to self: in order to build, use the Maven window in IntelliJ to trigger `compile` and then `package`.


enterprise/kerberos-login-module/src/main/java/com.hazelcast.security/KerberosLoginModule.java
```diff
@@ -0,0 +1,196 @@
...
+    public boolean onLogin() throws LoginException {
+
+        // get the token
+        byte[] token = getToken();
+
+        // accept the token
+        GSSContext gssContext = acceptToken(token);
+
+        // validate access (throws if access is denied)
+        validateAccess(gssContext);
```
kwart: I would rename the method to a more descriptive name here and wrap it into the `skipRole` condition.
E.g.

```java
if (! isSkipRole()) {
  addGroupSidsFromMsPac(token);
}
```

enterprise/kerberos-login-module/src/main/java/com.hazelcast.security/KerberosLoginModule.java
```diff
@@ -0,0 +1,196 @@
...
+                pacAuthzEntry = null;
+            }
+
+            if (authzEntry == null) {
+                throw new FailedLoginException("Internal error (no AD_IF_RELEVANT entry).");
+            }
+
+            if (pacAuthzEntry == null) {
+                throw new FailedLoginException("Internal error (no AD-WIN2K-PAC entry within the AD-IF-RELEVANT entry).");
+            }
```
kwart: I think we don't need these checks. The missing role means no permission mapped, so we should be safe. If you still want to have the checks included, I would suggest renaming the project from `kerberos-login-module` to `active-directory-login-module`.

enterprise/kerberos-login-module/pom.xml
```diff
@@ -0,0 +1,69 @@
...
+          <artifactId>kerberos-login-module</artifactId>
+          <version>${hazelcast.version}</version>
+          <configuration>
+            <archive>
+              <manifestFile>src/main/resources/META-INF/MANIFEST.MF</manifestFile>
+            </archive>
+          </configuration>
+        </plugin>
+      </plugins>
+    </build>
```
kwart: Probably we don't need this part. Od do you implement some Maven plugin within the code?

enterprise/kerberos-login-module/pom.xml
```diff
@@ -0,0 +1,69 @@
...
+            <version>${project.version}</version>
+        </dependency>
+        
+        <dependency>
+          <groupId>KerberosUtils</groupId>
+          <artifactId>KerberosUtils</artifactId>
+          <version>0.0</version>
+          <scope>system</scope>
+          <systemPath>${project.basedir}/lib/KerberosUtils.jar</systemPath>
+        </dependency>
```
kwart: We should avoid system dependencies.

enterprise/kerberos-login-module/src/main/java/com.hazelcast.security/KerberosLoginModule.java
```diff
@@ -0,0 +1,196 @@
...
+package com.hazelcast.security;
+
+import javax.security.auth.callback.Callback;
+import javax.security.auth.callback.UnsupportedCallbackException;
+import javax.security.auth.login.FailedLoginException;
+import javax.security.auth.login.LoginException;
+
+import com.hazelcast.internal.util.StringUtil;
+import com.sun.security.jgss.ExtendedGSSContext;
+import com.sun.security.jgss.InquireType;
```
kwart: The README should mention the LoginModule requires Oracle/OpenJDK JVM.

enterprise/kerberos-login-module/pom.xml
zpqrtbnk: Realized we don't need it indeed, removing.

enterprise/kerberos-login-module/src/main/java/com.hazelcast.security/KerberosLoginModule.java
zpqrtbnk: Adding to README.

enterprise/kerberos-login-module/src/main/java/com.hazelcast.security/KerberosLoginModule.java
zpqrtbnk: Makes sense. Fixing: if we cannot find the entries, we just don't map any more.

enterprise/kerberos-login-module/src/main/java/com.hazelcast.security/KerberosLoginModule.java
zpqrtbnk: Understood. Refactoring.

enterprise/kerberos-login-module/pom.xml
zpqrtbnk: Understood. I am using classe for [this library](https://github.com/jcmturner/java-kerberos-utils) which is published on GitHub but does not seem to be available on a Maven repository of some sort - what would be the best solution here? Copy & embed the code in our module? Or?

zpqrtbnk: Have experimented with JaasLounge and other libs - all libs are quite old, which is ... weird ... anyways the simplest solution for now seems to be to embed the bare minimum of the KerberosUtils lib I was using. So, have updated the PR.

enterprise/kerberos-login-module/README.md
```diff
@@ -0,0 +1,224 @@
...
+
+http://hazelcast.com/hazelcast-enterprise-trial/
+
+## Introduction
+
+Demonstrates how to create a Kerberos Login Module for Hazelcast.
+
+The project builds and produces the `kerberos-login-module.jar` artifact. It relies on the [Java Kerberos Utilities library](https://github.com/jcmturner/java-kerberos-utils) which is available under the Apache License 2.0. The library is included in the project's `lib` directory, and embedded in the produced artifact.
+
+The module relies on Java GSS and therefore requires Oracle/OpenJDK JVM.
```
kwart: The JGSS is available in IBM JDKs too, the problem is in using private `com.sun.security` and `sun.security` packages. So I would choose another wording here. E.g.

> The module relies on internal Java GSS implementation classes in `com.sun.security` and `sun.security` packages and therefore requires Oracle/OpenJDK JVM.

zpqrtbnk: Fixing

kwart:
The build fails on checkstyle errors. Probably it will need some exclusion rules due to the imported source files.

mmedenjak:
@zpqrtbnk @kwart can we merge this PR or progress on it somehow?

zpqrtbnk:
This was an experimental PR while we developed Kerberos support on top of IMDG 4.0, but now a different and more stable / official solution is available as part of IMDG 4.1 which officially supports Kerberos, so this PR is now moot. Closing.

PR #408, opened 2020-03-12
Remove stale hazelcast-integration/javascript sample

kwart:
Removing outdated nodejs sample which doesn't use the official nodejs client.

The up-to-date samples are in the `clients/nodejs`.


kwart:
Thanks for reviews, guys.

PR #428, opened 2020-05-29
Refresh Spring Hibernate 2nd Level Cache

pivovarit:


pivovarit:
That's really weird. If I remove the `hazelcast.xml` config file, there's no `HazelcastInstance` bean created, but a node starts anyway 🤯

pivovarit:
So, I think that for the sake of this guide, we need to assume that users want to use Hazelcast for 2LC exclusively. With this in mind, we can simply turn-off the Spring autoconfiguration:

```
@SpringBootApplication(exclude = HazelcastAutoConfiguration.class) // in order to avoid autoconfiguring an extra Hibernate instance
public class SpringHibernate2lcApplication {

    public static void main(String[] args) {
        SpringApplication.run(SpringHibernate2lcApplication.class, args);
    }
}
```

And then include a note on what to do you if you actually want to expose it as `HazelcastInstance` bean.

enozcan:
>  we need to assume that users want to use Hazelcast for 2LC exclusively.....And then include a note on what to do you if you actually want to expose it as HazelcastInstance bean.

Yes, that looks reasonable.

hazelcast-integration/spring-hibernate-2ndlevel-cache/src/main/java/com/hazelcast/hibernate/springhibernate2lc/SpringHibernate2lcApplication.java
```diff
@@ -2,10 +2,11 @@
 
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.autoconfigure.hazelcast.HazelcastAutoConfiguration;
 import org.springframework.scheduling.annotation.EnableScheduling;
 
-@EnableScheduling
-@SpringBootApplication
+@EnableScheduling // needed just for the sake of SecondLevelCacheVisualizer
+@SpringBootApplication(exclude = HazelcastAutoConfiguration.class) // in order to avoid autoconfiguring an extra Hibernate instance
```
enozcan, line 9: Hibernate instance -> Hazelcast instance

pivovarit: 🤦‍♀️

PR #438, opened 2020-06-16
Enhance Spring Boot - Hibernate 2LC samples with native client support

pivovarit:


hazelcast-integration/spring-hibernate-2ndlevel-cache/README.md
```diff
@@ -6,6 +6,10 @@ You can spin-up a PostgreSQL instance easily using Docker:
 
 `docker run --name 2lc-postgres --publish 5432:5432 -e POSTGRES_PASSWORD=mysecretpassword -d postgres:13`
 
+If you want to use Hazelcast client, you can start an IMDG instance in a Docker container easily:
+
+`docker run -p 5701:5701 hazelcast/hazelcast:4.0.1`
```
mesutcelik, line 11: It would be better not to use hardcoded version.

pivovarit: I can get rid of that, but the downside is that people would need to guess which version works with the guide. What about adding a note that there might be a newer version available?

mesutcelik: What about `docker run -p 5701:5701 hazelcast/hazelcast` so people would pull the latest always?

pivovarit: That's the same problem. One extra parameter that sneaks into the equation determining if the guide works or not. However, we could just add a note that they should carefully choose the supported version?

mesutcelik: I am ok with that.

PR #442, opened 2020-06-26
Refresh docker-compose code sample

leszko:


leszko:
> 👍 Are you planning to move it to guides repo?

When the Guide site is ready, then we'll probably walk through all integration code samples and choose those which should be moved to guides.

PR #445, opened 2020-07-09
Remove springboot cache code sample

leszko:
Moved to guides: https://github.com/hazelcast-guides/springboot-caching


mesutcelik:
What do you think about keeping the Readme and link back to new guide?

leszko:
> What do you think about keeping the Readme and link back to new guide?

Let's do it for some time and then remove it completely.

PR #448, opened 2020-07-20
Clean up spring-* code samples

leszko:
Clean up spring-* code samples. I kept only one code sample `spring-configuration`. Fixed it and updated it to the latest spring version. Now it works fine. It presents how to configuration Hazelcast using Spring XML configuration.

I removed all other code samples, because:
- they don't work and nobody complains (so I guess they are not used)
- they present concepts for which we have identical code samples for spring boot (and spring boot is way more popular/trending)
- they are old and you can find much better resources on the Internet (e.g. `spring-security`, it's way better described in the Spring Documentation)

CC: @blazember @mmedenjak


leszko:
> https://github.com/hazelcast/hazelcast-code-samples/blob/master/hazelcast-integration/README.md
> 
> Can we update this as well?

Updated

hazelcast-integration/spring-configuration/src/main/java/com/hazelcast/springconfiguration/HazelcastDataTypes.java
```diff
@@ -181,17 +179,6 @@ private static void executeSemaphore() {
...
-        System.out.println("### Lock Execution Started... ###");
-        FencedLock lock = context.getBean("lock", FencedLock.class);
-        lock.lock();
-        System.out.println("lock() call...");
-        System.out.println("is locked by current thread: " + lock.isLockedByCurrentThread());
-        lock.unlock();
-        System.out.println("unlock() call...");
-        System.out.println("is locked: " + lock.isLocked());
-    }
-
```
blazember, line 194: Is there a reason you removed `FencedLock` while keeping the others?

leszko: My understanding is that this part has changed since the introduction of CP subsystem.

PR #451, opened 2020-07-21
Remove eureka partition groups and sessions analysis code samples

leszko:
Reasoning:

- these code samples are rather use-cases, not simple code sample; they are hard to maintain and not very useful to the general public
- both of them have the copy of the content as a blog post or whitepaper:
  - https://hazelcast.com/resources/data-safety-eureka/
  - https://hazelcast.com/blog/session-analysis-for-an-online-shop/
- Eureka partition groups code sample actually shows the way we don't recommend for using Eureka
  - the recommended way is to use `hazelcast-eureka-one` plugin
  - this code sample uses custom SPI implementation


neilstevenson:
These examples still have limited value for those use-cases. While there may be better approaches, showing other approaches could be useful.

Can we add a comment to the top of the `README.md` for each, something like
```
OBSOLETE. 
*NOT MAINTAINED*
*DO NOT USE*
Retained in case it is useful backgriound, but the better approach is to ....
```

Then our only effort is to ensure it compiles, which presumably won't be too difficult if the versions are locked.

mmedenjak:
I'm for removing and/or replacing instead of keeping obsolete code samples.

neilstevenson:
If we delete, search engines with stale caches will direct people to broken links. If the code is still there and tagged as _OBSOLETE_ they can decide what to do.

leszko:
I'm also on the side of removing obsolete things. Anyway, let's wait until next week for @mesutcelik. Maybe he wants to keep them.

I think the old blog post is the way to keep old and not maintained information.

mesutcelik:
+1 removing it. I checked blog post and pdf and they do not have a link to this repo too. Google Search will update its indexes over time and will not send traffic once we remove the content.

I think we should either keep maintaining every sample actively as products or just remove them.

leszko:
Ok, then I guess we should remove them.

neilstevenson:
Just to confirm, I am not in favour of removing obsolete content. My opinion is to mark it obsolete and leave it in place.

Someone somewhere will be running old code, and other search engines may find it.

leszko:
Neil, I see your point. But I guess keeping obsolete content in the Code Samples repo is not a good idea (as also pointed by @mmedenjak and @mesutcelik). The thing is that we should decide on what content is the most important and keep it up-to-date in Code Samples and [Hazelcast Guides](https://guides.hazelcast.org/home/). Otherwise, we risk having a lot of garbage which doesn't work, which may make someone thing the whole project is garbage.

PR #464, opened 2020-12-01
Replace Guava's Joiner with Java's String.join()

pivovarit:


leszko (approved):

lazerion:
@pivovarit it also needs a backport to 4.1.z line.

PR #472, opened 2021-03-23
MTS Hazelcast evaluation and prototyping - first commit

andrewjessop:
- Trial some predicate/aggregate logic
- Execute an insert/query performance test with 1m/2m/4m rows on a local cluster of 3


andrewjessop:
Incorrectly raised this against the upstream repo, instead of my forked version.

Sorry for the notifications.

PR #488, opened 2021-09-08
Adds code samples for compact serialization [API-367]

sancar:
And SQL usages of compact serialization

### DIFF
--- a/serialization/compact/src/main/java/CompactFullConfig.java
+++ b/serialization/compact/src/main/java/CompactFullConfig.java
@@ -19,7 +19,7 @@ public class CompactFullConfig {
         //This config is needed only during BETA phase.
         CompactSerializationConfig compactSerializationConfig = config.getSerializationConfig().getCompactSerializationConfig();
         compactSerializationConfig.setEnabled(true);
-        //Here we register a typename and a serializer against the PersonDTO class so that any client from any language
+        //Here we register a typename `person` and a serializer against the PersonDTO class so that any client from any language
         //can use same typename abd field names to match
         compactSerializationConfig.register(PersonDTO.class, "person", new CompactSerializer<PersonDTO>() {
             @Override
--- a/serialization/compact/src/main/java/CompactFullConfig.java
+++ b/serialization/compact/src/main/java/CompactFullConfig.java
@@ -6,7 +6,6 @@ import com.hazelcast.nio.serialization.compact.CompactReader;
 import com.hazelcast.nio.serialization.compact.CompactSerializer;
 import com.hazelcast.nio.serialization.compact.CompactWriter;
 
-import java.io.IOException;
 import java.util.Map;
 
 /**
@@ -21,7 +20,7 @@ public class CompactFullConfig {
         CompactSerializationConfig compactSerializationConfig = config.getSerializationConfig().getCompactSerializationConfig();
         compactSerializationConfig.setEnabled(true);
         //Here we register a typename and a serializer against the PersonDTO class so that any client from any language
-        // can use same typename abd field names to match 
+        //can use same typename abd field names to match
         compactSerializationConfig.register(PersonDTO.class, "person", new CompactSerializer<PersonDTO>() {
             @Override
             public PersonDTO read(CompactReader compactReader) {
--- a/serialization/compact/src/main/java/CompactFullConfig.java
+++ b/serialization/compact/src/main/java/CompactFullConfig.java
@@ -20,7 +20,7 @@ public class CompactFullConfig {
         CompactSerializationConfig compactSerializationConfig = config.getSerializationConfig().getCompactSerializationConfig();
         compactSerializationConfig.setEnabled(true);
         //Here we register a typename `person` and a serializer against the PersonDTO class so that any client from any language
-        //can use same typename abd field names to match
+        //can use same typename and field names to match
         compactSerializationConfig.register(PersonDTO.class, "person", new CompactSerializer<PersonDTO>() {
             @Override
             public PersonDTO read(CompactReader compactReader) {
--- a/serialization/compact/src/main/java/CompactFullConfig.java
+++ b/serialization/compact/src/main/java/CompactFullConfig.java
@@ -1,0 +1,51 @@
+import com.hazelcast.config.CompactSerializationConfig;
+import com.hazelcast.config.Config;
+import com.hazelcast.core.Hazelcast;
+import com.hazelcast.core.HazelcastInstance;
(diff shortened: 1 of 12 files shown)


serialization/compact/src/main/java/CompactFullConfig.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+public class CompactFullConfig {
+
+    public static void main(String[] args) {
+        Config config = new Config();
+        //This config is needed only during BETA phase.
+        CompactSerializationConfig compactSerializationConfig = config.getSerializationConfig().getCompactSerializationConfig();
+        compactSerializationConfig.setEnabled(true);
+        //Here we register a typename and a serializer against the PersonDTO class so that any client from any language
+        // can use same typename abd field names to match
```
ihsandemir:
```suggestion
        // can use same typename and field names to match 
```

serialization/compact/src/main/java/CompactFullConfig.java
```diff
@@ -0,0 +1,51 @@
...
+ * different languages needs to be used.
+ */
+public class CompactFullConfig {
+
+    public static void main(String[] args) {
+        Config config = new Config();
+        //This config is needed only during BETA phase.
+        CompactSerializationConfig compactSerializationConfig = config.getSerializationConfig().getCompactSerializationConfig();
+        compactSerializationConfig.setEnabled(true);
+        //Here we register a typename and a serializer against the PersonDTO class so that any client from any language
```
ihsandemir, line 22:
```suggestion
        //Here we register a typename `person` and a serializer against the PersonDTO class so that any client from any language
```

serialization/compact/src/main/java/CompactFullConfig.java
```diff
@@ -0,0 +1,51 @@
...
+import com.hazelcast.nio.serialization.compact.CompactWriter;
+
+import java.io.IOException;
+import java.util.Map;
+
+/**
+ * Demonstrates how to use Compact(BETA) serialization from java in a polyglot setup where hazelcast clients written in
+ * different languages needs to be used.
+ */
+public class CompactFullConfig {
```
ihsandemir, line 15:
```suggestion
public class CompactProgrammaticConfig {
```

serialization/compact/src/main/java/CompactZeroConfig.java
```diff
@@ -0,0 +1,32 @@
...
+ * If all hazelcast clients and members are java, then you can use Compact format without any config.
+ */
+public class CompactZeroConfig {
+
+    public static void main(String[] args) {
+        Config config = new Config();
+        //This config is needed only during BETA phase.
+        CompactSerializationConfig compactSerializationConfig = config.getSerializationConfig().getCompactSerializationConfig();
+        compactSerializationConfig.setEnabled(true);
+        //Note that PersonDTO does not implement Serializable. In such case, following config should be added
```
ihsandemir, line 19: This detail may be confusing in this example. May be you can add another example just for this.

sql/compact/src/main/java/NoClassExample.java
```diff
@@ -0,0 +1,55 @@
...
+    public static void main(String[] args) throws InterruptedException {
+        Config config = new Config();
+        config.getJetConfig().setEnabled(true);
+        config.getSerializationConfig().getCompactSerializationConfig().setEnabled(true);
+        HazelcastInstance hazelcast = Hazelcast.newHazelcastInstance(config);
+        IMap<Integer, GenericRecord> myMap = hazelcast.getMap("myMap");
+
+        SqlService sqlService = hazelcast.getSql();
+
+        sqlService.execute("CREATE MAPPING myMap ( " +
```
ihsandemir: Can we add a comment on what this statement does at the background briefly. "It creates a table named myMap with columns listed, etc." Indicate the meaning of the `valueCompactTypeName ` and `valueFormat ` fields.

sql/compact/src/main/java/NoClassExample.java
```diff
@@ -0,0 +1,55 @@
...
+                "    'keyFormat' = 'int', " +
+                "    'valueFormat' = 'compact', " +
+                "    'valueCompactTypeName' = 'person' )");
+
+        sqlService.execute("INSERT INTO myMap (__key, name, surname, id) VALUES (1, ?, ?, 1)", "John", "Stone");
+
+        //Query map with sql
+        SqlResult sqlRows = sqlService.execute("SELECT * FROM myMap");
+        for (SqlRow sqlRow : sqlRows) {
+            System.out.println(sqlRow);
```
ihsandemir, line 43: Can you put a comment what this outputs?

sql/compact/src/main/java/WithExistingClassExample.java
```diff
@@ -0,0 +1,57 @@
...
+        compactSerializationConfig.setEnabled(true);
+        config.getJetConfig().setEnabled(true);
+        HazelcastInstance hazelcast = Hazelcast.newHazelcastInstance(config);
+
+        IMap<Integer, Person> myMap = hazelcast.getMap("myMap");
+        myMap.put(1, new Person(1, "John", "Stone"));
+
+        SqlService sqlService = hazelcast.getSql();
+
+        sqlService.execute("CREATE MAPPING myMap ( " +
```
ihsandemir: A comment saying that this reuses existing schema since the fields are the same.

serialization/compact/src/main/java/CompactZeroConfig.java
sancar: I am having trouble naming the examples. Since this is one line addition, I felt like this is a good point to add.

serialization/compact/src/main/java/CompactFullConfig.java
sancar: I want this to be not example of ProgrammaticConfig. This is Zero Config vs ? . And I think Full Config is a good name for ?
What do you think ?

serialization/compact/src/main/java/CompactZeroConfig.java
ihsandemir: Having a separate example would make this distinction more visible to the user but I am also OK with this then.

ihsandemir, line 19:
```suggestion
        //If PersonDTO were implementing Java Serializable interface, then you would need to add the following config do that the object is serialized using the Compact serialization. Otherwise, it defaults to Java serialization.
```

ihsandemir (approved):

PR #489, opened 2021-09-14
Enabled Jet in some code samples

olukas:
Some Jet code samples don't work because Jet is disabled there.


jet/job-management/src/main/resources/hazelcast.yaml
```diff
@@ -13,4 +13,7 @@ hazelcast:
     source:
       event-journal:
         enabled: true
-        capacity: 100000
\ No newline at end of file
+        capacity: 100000
+
+  jet:
+    enabled: true
```
ufukyilmaz, line 19: we can add a newline here

ufukyilmaz (approved):

PR #529, opened 2022-04-20
Allow for running samples against the external Kafka HZ-949

ldziedziul:
Related to https://hazelcast.atlassian.net/browse/HZ-949

Added a new system property `use.embedded.kafka` to disable the embedded Kafka and thus enable usage of the external Kafka with our code samples. By default the code samples will spawn the embedded Kafka. The same goes with `-Duse.embedded.kafka=true`

With `-Duse.embedded.kafka=false` code sample will expect that Kafka is running on `localhost:9092` (broker) and `localhost:8081` (schema registry)

## Testing 
Change directory to `jet/kafka`

On system without Kafka run 
```shell 
mvn compile exec:java -Dexec.cleanupDaemonThreads=false \
 -Dexec.mainClass=com.hazelcast.samples.jet.kafka.KafkaSource -Duse.embedded.kafka=true
``` 
It should create an embedded server and pass the build

Then run 
```shell
mvn compile exec:java -Dexec.cleanupDaemonThreads=false \
-Dexec.mainClass=com.hazelcast.samples.jet.kafka.KafkaSource -Duse.embedded.kafka=false
```
It should NOT create an embedded server and loop with messages like ` Broker may not be available.` 

Start Kafka in docker:
```shell
curl --silent --output docker-compose.yml \
  https://raw.githubusercontent.com/confluentinc/cp-all-in-one/7.1.1-post/cp-all-in-one/docker-compose.yml
docker-compose up
```
wait for initialization and then run:
```shell
mvn compile exec:java -Dexec.cleanupDaemonThreads=false \
-Dexec.mainClass=com.hazelcast.samples.jet.kafka.KafkaSource -Duse.embedded.kafka=false
```
It should NOT create an embedded server and pass the build

### DIFF
--- a/jet/kafka/src/main/java/com/hazelcast/samples/jet/kafka/KafkaSink.java
+++ b/jet/kafka/src/main/java/com/hazelcast/samples/jet/kafka/KafkaSink.java
@@ -55,9 +55,9 @@ public class KafkaSink {
 
     private static final int MESSAGE_COUNT = 50_000;
     private static final String BOOTSTRAP_SERVERS = "localhost:9092";
-
     private static final String SOURCE_NAME = "source";
     private static final String SINK_TOPIC_NAME = "t1";
+    private static final boolean USE_EMBEDDED_KAFKA = Boolean.parseBoolean(System.getProperty("use.embedded.kafka", "true"));
 
     private EmbeddedZookeeper zkServer;
     private KafkaServer kafkaServer;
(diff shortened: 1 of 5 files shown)


jet/kafka/src/main/java/com/hazelcast/samples/jet/kafka/KafkaSink.java
```diff
@@ -141,8 +146,8 @@ public <K, V> KafkaConsumer<K, V> createConsumer(
         return consumer;
     }
 
-    // Creates an embedded zookeeper server and a kafka broker
     private void createKafkaCluster() throws IOException {
+        System.out.println("Creating an embedded zookeeper server and a kafka broker");
```
frant-hartm, line 150: Use `log` instead?

frant-hartm (approved):

PR #533, opened 2022-04-26
Use String values in Kafka samples to make them readable in Confluent Control Center UI

ldziedziul:
Related to https://github.com/hazelcast/hazelcast-code-samples/pull/529
Related to https://hazelcast.atlassian.net/browse/HZ-949

Use String values in Kafka samples to make them readable in Confluent Control Center UI

Before:
image001

After:
image


frant-hartm (approved):
Just an idea, not really a blocker - maybe we could use the Integer for the key, so we have different types for the key and value. Would that work with the Confluent control center?

ldziedziul:
> Just an idea, not really a blocker - maybe we could use the Integer for the key, so we have different types for the key and value. Would that work with the Confluent control center?

It seems that items have to have unique keys, I think it would complicate the sample too much.

PR #540, opened 2022-06-20
CDC real time analytics demo

TomaszGaweda:
Simple demo showing how user can use CDC for real-time analytics without changing anything in source application.

Requires https://github.com/hazelcast/hazelcast/pull/21536 to be installed in local repo/merged before build (relies on new functions).


TomaszGaweda:
run-lab-run

jet/cdc/src/main/java/com/hazelcast/samples/jet/cdc/Order.java
```diff
@@ -0,0 +1,71 @@
+package com.hazelcast.samples.jet.cdc;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.io.Serializable;
+import java.util.Date;
+import java.util.Objects;
+
+@SuppressWarnings("checkstyle:VisibilityModifier")
+public class Order implements Serializable {
```
ldziedziul, line 10: Maybe?
```suggestion
public record Order implements Serializable {
```

jet/cdc/src/main/java/com/hazelcast/samples/jet/cdc/CdcRealTimeAnalysisWithParallelSnapshotDemo.java
```diff
@@ -0,0 +1,168 @@
...
+                    if (record.event() instanceof Customer customer) {
+                        state.setCustomerFirstName(customer.firstName);
+                        state.setCustomerLastName(customer.lastName);
+                        state.setCustomerId(customer.id);
+
+                    } else  {
+                        var order = (Order) record.event();
+                        var operation = record.operation();
+                        if (operation == Operation.SYNC || operation == INSERT) {
+                            if (!state.processedOrderIds.add(order.getId())) return state;
```
ldziedziul, line 127: why this condition?

jet/cdc/src/main/java/com/hazelcast/samples/jet/cdc/CdcRealTimeAnalysisDemo.java
```diff
@@ -0,0 +1,142 @@
...
+ * Demonstrates a simple reporting pipeline, which will be attached to existing system without the need of any change
+ * in the upstream system and will update some analytics based on actions done by customers.
+ *
+ * To run this example you have to have PostgreSQL database set up. You can quickly do this, by running in your terminal:
+ * <pre>
+ *     docker run -d --rm --name cdc-postgres -p 5432:5432 \
+ *      -e POSTGRES_DB=postgres -e POSTGRES_USER=postgres \
+ *      -e POSTGRES_PASSWORD=cdcpwd1337 debezium/example-postgres:1.9.3.Final
+ * </pre>
+ *
```
ldziedziul, line 32: I was a bit surprised that I couldn't find the place where we populate the DB. Maybe it would be worth mentioning that `debezium/example-postgres` is filled with sample data or even adding a code that changes DB data to see that the integration works in the real time.

ldziedziul (commented):
Nice work, left some comments

jet/cdc/src/main/java/com/hazelcast/samples/jet/cdc/Order.java
TomaszGaweda: racja! Customer też od razu poprawiłem :)

TomaszGaweda: Indeed! I've corrected also Customer entityr

ldziedziul (approved):
Great job!

jet/cdc/src/main/java/com/hazelcast/samples/jet/cdc/CdcRealTimeAnalysisDemo.java
```diff
@@ -0,0 +1,144 @@
...
+                .setEnabled(true);
+
+        HazelcastInstance hz = Hazelcast.newHazelcastInstance(config);
+        JetService jet = hz.getJet();
+        // trigger initialization
+        hz.getMap("CustomerStatsReport");
+
+        hz.getSql().execute(
+          """
+                  create mapping CustomerStatsReport (
```
frant-hartm, line 56: I would make all SQL keywords (`CREATE MAPPING`, `INT`, `VARCHAR`, `TYPE`, ..) uppercase, I think that's our convention in our docs.

jet/cdc/src/main/java/com/hazelcast/samples/jet/cdc/CdcRealTimeAnalysisWithParallelSnapshotDemo.java
```diff
@@ -0,0 +1,173 @@
...
+        var cdc = pipeline.readFrom(
+                        postgres("inventory-db")
+                                .setDatabaseName("postgres")
+                                .setSchemaWhitelist("inventory")
+                                .setDatabaseAddress("localhost")
+                                .setDatabaseUser("postgres")
+                                .setDatabasePassword("cdcpwd1337")
+                                .setTableWhitelist("inventory.orders", "inventory.customers")
+                                .setDatabasePort(5432)
+                                .setCustomProperty("snapshot.mode", "initial")
```
frant-hartm, line 100: If I understand correctly this will run initial snapshot, which you also do in parallel with the jdbc source. Did you mean to use other value here?

jet/cdc/src/main/java/com/hazelcast/samples/jet/cdc/CdcRealTimeAnalysisDemo.java
```diff
@@ -0,0 +1,144 @@
...
+                            state.setItemsTotal(state.getItemsTotal() + order.quantity());
+                            state.setOrdersTotal(state.getOrdersTotal() + 1);
+                            state.setItemsAvg(state.getItemsTotal() * 1.0d / state.getOrdersTotal());
+                        } else if (operation == DELETE) {
+                            state.setItemsTotal(state.getItemsTotal() - order.quantity());
+                            state.setOrdersTotal(state.getOrdersTotal() - 1);
+                            state.setItemsAvg(state.getItemsTotal() * 1.0d / state.getOrdersTotal());
+                        }
+                    }
+                    return state;
```
frant-hartm, line 114: I think you should return a copy here, otherwise, you will overwrite the data in the instance, which gets sent further down the pipeline, the changes might, or might not be visible.

jet/cdc/src/main/java/com/hazelcast/samples/jet/cdc/CdcRealTimeAnalysisDemo.java
```diff
@@ -0,0 +1,144 @@
...
+                                .setDatabasePassword("cdcpwd1337")
+                                .setDatabasePort(5432)
+                                .setTableWhitelist("inventory.customers", "inventory.orders")
+                                .build())
+                .withoutTimestamps()
+                .flatMap(record -> {
+                    if (isOrder(record) && record.operation() == Operation.UPDATE) {
+                        return traverseItems(
+                                orderEvent(record.newValue().toObject(Order.class), INSERT),
+                                orderEvent(record.oldValue().toObject(Order.class), DELETE)
```
frant-hartm, line 86: Shouldn't you traverse this in reversed order? Otherwise, you might temporarily get larger result with both old and new version of the order.

jet/cdc/src/main/java/com/hazelcast/samples/jet/cdc/CdcRealTimeAnalysisDemo.java
```diff
@@ -0,0 +1,144 @@
...
+                        state.setCustomerId(customer.id());
+
+                    } else  {
+                        var order = (Order) record.event();
+                        var operation = record.operation();
+                        if (operation == Operation.SYNC || operation == INSERT) {
+                            state.setCustomerId(order.purchaser());
+                            state.setItemsTotal(state.getItemsTotal() + order.quantity());
+                            state.setOrdersTotal(state.getOrdersTotal() + 1);
+                            state.setItemsAvg(state.getItemsTotal() * 1.0d / state.getOrdersTotal());
```
frant-hartm, line 107: You can move this to the `CustomerStatsReport`?

jet/cdc/src/main/java/com/hazelcast/samples/jet/cdc/CdcRealTimeAnalysisWithParallelSnapshotDemo.java
```diff
@@ -0,0 +1,173 @@
...
+                            state.setItemsTotal(state.getItemsTotal() + order.quantity());
+                            state.setOrdersTotal(state.getOrdersTotal() + 1);
+                            state.setItemsAvg(state.getItemsTotal() * 1.0d / state.getOrdersTotal());
+                        } else if (operation == DELETE) {
+                            state.setItemsTotal(state.getItemsTotal() - order.quantity());
+                            state.setOrdersTotal(state.getOrdersTotal() - 1);
+                            state.setItemsAvg(state.getItemsTotal() * 1.0d / state.getOrdersTotal());
+                        }
+                    }
+                    return state;
```
frant-hartm, line 143: Copy needed here.

jet/cdc/src/main/java/com/hazelcast/samples/jet/cdc/Order.java
```diff
@@ -0,0 +1,17 @@
...
+
+public record Order (
+        int id,
+        @JsonProperty("order_date")
+        Date orderDate,
+        int purchaser,
+        int quantity,
+        @JsonProperty("product_id")
+        int productId
+) implements Serializable {
```
frant-hartm, line 16: Is is possible to use compact for order instead of Serializable?

frant-hartm (changes requested):

jet/cdc/src/main/java/com/hazelcast/samples/jet/cdc/CdcRealTimeAnalysisWithParallelSnapshotDemo.java
TomaszGaweda: This is a tricky one. We are streaming the Debezium's snapshot data, but also reading them ourselves. The "slower" data will be discarded later in the mapStateful. This will make reading faster, but still not optimal (for optimal reading we would need to modify Debezium).

It would be better to have separate values for each table, but debezium doesn't support that and we can create only 1 source for one database (otherwise Debezium fails).

jet/cdc/src/main/java/com/hazelcast/samples/jet/cdc/Order.java
TomaszGaweda: yes, in fact I directly enable compact at the beginning. I will drop implements Serializable, so it's not confusing

jet/cdc/src/main/java/com/hazelcast/samples/jet/cdc/CdcRealTimeAnalysisWithParallelSnapshotDemo.java
TomaszGaweda: Done

jet/cdc/src/main/java/com/hazelcast/samples/jet/cdc/CdcRealTimeAnalysisDemo.java
TomaszGaweda: +1, looks much cleaner. Done

jet/cdc/src/main/java/com/hazelcast/samples/jet/cdc/CdcRealTimeAnalysisDemo.java
TomaszGaweda: Yeah, that's a good idea. In general the timing shouldn't really be an issue, but worth changing, just in case :)

frant-hartm (changes requested):
I would like to see some README.md in the cdc folder to describe the demo - take what you have on the Javadoc and enhance it a bit.

I know there is more info in the blog post, but example should be self-contained.

jet/cdc/src/main/java/com/hazelcast/samples/jet/cdc/CustomerStatsReport.java
```diff
@@ -0,0 +1,136 @@
...
+        itemsTotal = itemsTotal - order.quantity();
+        ordersTotal = ordersTotal - 1;
+        itemsAvg = itemsTotal * 1.0d / ordersTotal;
+    }
+
+    public synchronized void updateCustomerData(Customer customer) {
+        customerFirstName = customer.firstName();
+        customerLastName = customer.lastName();
+        customerId = customer.id();
+    }
```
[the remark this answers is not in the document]
TomaszGaweda: Nice catch, indeed. I went copy-paste mode unfortunatelly :P

TomaszGaweda:
Done - I've added README with some context, more detailed explanations are - as always - in JavaDoc of concrete example

frant-hartm (approved):

PR #546, opened 2022-09-02
Add support for Compact Serialization to the serialization benchmark.

myawnhc:


mdumandag:
@myawnhc I think we should add compact serializers for those classes. Right now, they are using the reflective serializer, and it has a considerable overhead, compared to the manually written serializers (setting fields with reflection vs a normal write to a field etc.)

PR #560, opened 2023-01-26
Add code sample to simulate max idle with ttl [HZ-1981]

ramizdundar:
Max idle behavior can be simulated via ttl usage. This PR aims to add a
code sample related to that. See `README` and `AbstractMaxIdleSimulator`
javadoc for more information.


ramizdundar:
Dear reviewers, although I added migration listeners to not block/throw during migrations, I saw no practical difference using them. Please see `BouncingSingleMaxIdleSimulator#USE_NORMAL_GETS_DURING_MIGRATION` and try to switch it off. You can also add a logging inside `AbstractMaxIdleSimulator#simulate()` for loop. You will see that even if we use simulating get, `map.executeOnKey()` will not block or throw during migrations.

enterprise/wan-replication/simulating-max-idle/README.md
```diff
@@ -0,0 +1,53 @@
...
+As a workaround for this issue, if you don't care about:
+
+ - Entries may live longer than configured max idle or,
+ - You calculate the expiration yourself
+
+You can use an `EntryProcessor`.
+
+## Detail
+
+Ideally to solve the issue, we can do write after each read. But that
```
ahmetmircik, line 16:
```suggestion
Ideal solution is wan-replicating reads as in writes. But that
```

enterprise/wan-replication/simulating-max-idle/src/main/java/MultipleMaxIdleSimulator.java
```diff
@@ -0,0 +1,82 @@
...
+        private transient HazelcastInstance instance;
+        private transient volatile boolean readOnly = true;
+
+        @Override
+        public V process(Entry<K, V> entry) {
+            EntryView<K, V> entryView = instance.<K, V>getMap(MAP_NAME).getEntryView(entry.getKey());
+
+            if (entryView == null) {
+                return null;
+            } else if (Clock.currentTimeMillis() - entryView.getLastUpdateTime() > ttlToMaxIdle((int) entryView.getTtl() / 1000)) {
```
ahmetmircik, line 33: Maybe we can note that to have last-update-time, per-entry-stats of map-config should be enabled.

ahmetmircik (approved):

enterprise/wan-replication/simulating-max-idle/src/main/java/BouncingSingleMaxIdleSimulator.java
```diff
@@ -0,0 +1,126 @@
...
+        @Override
+        public void replicaMigrationFailed(ReplicaMigrationEvent replicaMigrationEvent) {
+
+        }
+    }
+
+    private static class MigrationListenerImplB implements MigrationListener {
+
+        @Override
+        public void migrationStarted(MigrationState migrationState) {
```
ahmetmircik, line 59: I think 2 problems here:

- I believe this listener is called asynchronously. So we can have migration-started event but in reality migration can be completed already. There is a migration interceptor to aware from migration synchronously but it is internal. To see the true affect of migration, migration should be slow. Otherwise it is hard to detect possible latencies reliably in a test.
- migrationStarted is a global event and it has no partition id in it, maybe number of parallel partition can be set to 1 to catch latency effect better.

enterprise/wan-replication/simulating-max-idle/src/main/java/MultipleMaxIdleSimulator.java
ramizdundar: Hmm with this PR: https://github.com/hazelcast/hazelcast/pull/23279/files per entry records aren't required.

enterprise/wan-replication/simulating-max-idle/README.md
ramizdundar: I was trying to explain that a very naive approach would be to call put after every get. I tried to explain that better https://github.com/hazelcast/hazelcast-code-samples/pull/560/commits/664e5544ce8a82b46e4878d5ed9c335cc0f2d1c7

enterprise/wan-replication/simulating-max-idle/src/main/java/MultipleMaxIdleSimulator.java
ramizdundar: https://github.com/hazelcast/hazelcast-code-samples/pull/560/commits/664e5544ce8a82b46e4878d5ed9c335cc0f2d1c7

vbekiaris (approved):

PR #562, opened 2023-02-09
Update simulate max idle code sample [HZ-1981]

ramizdundar:
This PR includes minor updates to comments. Also redundant volatile
field made non-volatile.

Related: https://github.com/hazelcast/hazelcast-code-samples/pull/560

### DIFF
--- a/enterprise/wan-replication/simulating-max-idle/src/main/java/SingleMaxIdleSimulator.java
+++ b/enterprise/wan-replication/simulating-max-idle/src/main/java/SingleMaxIdleSimulator.java
@@ -19,7 +19,7 @@ class SingleMaxIdleSimulator extends AbstractMaxIdleSimulator {
 
     private static class MaxIdleSimulatingGet<K, V> implements EntryProcessor<K, V, V>, HazelcastInstanceAware {
         private transient HazelcastInstance instance;
-        private transient volatile boolean readOnly = true;
+        private transient boolean readOnly = true;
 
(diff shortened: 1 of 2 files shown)


enterprise/wan-replication/simulating-max-idle/src/main/java/SingleMaxIdleSimulator.java
```diff
@@ -17,22 +17,31 @@ public static void main(String[] args) {
         simulator.simulate();
     }
 
+    // You may want to consider using IdentifiedDataSerializable instead
+    // of default java serialization for better performance. If you do so,
+    // EntryProcessor, key and value should extend IdentifiedDataSerializable.
```
vbekiaris, line 22: I see why you suggest IDS format for the EP, but why should key & value also be IDS format?

vbekiaris (approved):

enterprise/wan-replication/simulating-max-idle/src/main/java/SingleMaxIdleSimulator.java
ramizdundar: It was a leftover from a changed backup processor logic, which is reverted now. https://github.com/hazelcast/hazelcast-code-samples/pull/562/commits/c7a40cf81feef1f54f3d435378fe7086f548fc55

ahmetmircik (approved):

PR #580, opened 2023-06-22
Add code sample for Jet JobStatusListener [HZ-2187]

sumnerib:
This sample is based on existing class `JobSuspendResume`.

### DIFF
--- a/jet/job-management/src/main/java/com/hazelcast/samples/jet/jobmanagement/JobSuspendResumeListener.java
+++ b/jet/job-management/src/main/java/com/hazelcast/samples/jet/jobmanagement/JobSuspendResumeListener.java
@@ -5,6 +5,7 @@ import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.jet.JetService;
 import com.hazelcast.jet.Job;
 import com.hazelcast.jet.config.JobConfig;
+import com.hazelcast.jet.core.JobStatus;
 import com.hazelcast.jet.pipeline.Pipeline;
 import com.hazelcast.jet.pipeline.Sinks;
 import com.hazelcast.jet.pipeline.Sources;
@@ -14,6 +15,8 @@ import java.util.concurrent.CancellationException;
 import static com.hazelcast.jet.pipeline.JournalInitialPosition.START_FROM_OLDEST;
 
 public class JobSuspendResumeListener {
+
+    private static int restartCount = -1;
 
     public static void main(String[] args) throws InterruptedException {
         // create two instances
@@ -33,12 +36,12 @@ public class JobSuspendResumeListener {
         Job job = jet1.newJob(p, jobConfig);
 
         // Register JobStatusListener
-        job.addStatusListener(event ->
-                System.out.printf("Job status changed: %s -> %s. User requested? %b%n",
-                    event.getPreviousStatus(),
-                    event.getNewStatus(),
-                    event.isUserRequested())
-        );
+        job.addStatusListener(event -> {
+            System.out.printf("Job status changed: %s -> %s. User requested? %b%n",
+                    event.getPreviousStatus(), event.getNewStatus(), event.isUserRequested());
+
+            if (event.getNewStatus() == JobStatus.RUNNING) restartCount++;
+        });
 
         // printing the job name
         System.out.println("Job '" + job.getName() + "' is submitted.");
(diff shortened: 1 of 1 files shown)


jet/job-management/src/main/java/com/hazelcast/samples/jet/jobmanagement/JobSuspendResumeListener.java
```diff
@@ -0,0 +1,72 @@
...
+                    event.isUserRequested())
+        );
+
+        // printing the job name
+        System.out.println("Job '" + job.getName() + "' is submitted.");
+
+        // we can suspend the job
+        Thread.sleep(1000);
+        System.out.println("Suspending the job...");
+        job.suspend();
```
burakgok, line 49: After suspending and resuming, we should wait for `SUSPENDED` and `RUNNING` statuses. Waiting one second may be enough for state transition, but the best practice is polling the status or listening to it.
```java
private static void waitForStatus(Job job, JobStatus status) throws InterruptedException {
    while (job.getStatus() != status) {
        Thread.sleep(100);
    }
}
```
By the way, it is not possible to observe a restart reliably by polling the status, but it can be observed by a listener since it is guaranteed to report `RUNNING` → `NOT_RUNNING` → `STARTING` → `RUNNING` transition. So, you might want to dedicate this sample (or a new one) to that usage, which is I think more interesting than logging the status. You can simply increment a count whenever you receive a `RUNNING` status and wait for a specific run count to infer a number of restarts. However, the unit tests show this particular usage and if they are accessible enough to users, you may skip this. If they aren't, we might consider encouraging users to read them because they usually demonstrate best practices.

burakgok (changes requested):

jet/job-management/src/main/java/com/hazelcast/samples/jet/jobmanagement/JobSuspendResumeListener.java
```diff
@@ -33,12 +36,12 @@ public static void main(String[] args) throws InterruptedException {
...
-                System.out.printf("Job status changed: %s -> %s. User requested? %b%n",
-                    event.getPreviousStatus(),
-                    event.getNewStatus(),
-                    event.isUserRequested())
-        );
+        job.addStatusListener(event -> {
+            System.out.printf("Job status changed: %s -> %s. User requested? %b%n",
+                    event.getPreviousStatus(), event.getNewStatus(), event.isUserRequested());
+
+            if (event.getNewStatus() == JobStatus.RUNNING) restartCount++;
```
burakgok, line 43: The initial `RUNNING` status might be missed and _restart count_ is not well-defined. What if I repeatedly restart a job before it even reaches the `RUNNING` status? In this implementation, the restart count would be zero although the job is not really running or pending (`NOT_RUNNING`).
1. Rename `restartCount` to `runCount` and initialize it to 1.
2. Add a separate status listener to increment run count. This listener should be added after ensuring that the job is running, i.e. `waitForStatus(job, JobStatus.RUNNING);`
3. Replace conditions `restartCount == n` with `runCount == n + 1`.

burakgok (changes requested):

jet/job-management/src/main/java/com/hazelcast/samples/jet/jobmanagement/JobSuspendResumeListener.java
sumnerib: I've implemented the suggested changes. Please keep in mind that the goal of the code samples is to provide an easily consumable and short demo of a specific feature for users. The examples are not meant to be complete solutions, and things like handling edge-cases can be considered out of scope.

jet/job-management/src/main/java/com/hazelcast/samples/jet/jobmanagement/JobSuspendResumeListener.java
```diff
@@ -0,0 +1,86 @@
...
+        // job name is optional
+        String jobName = "sampleJob";
+        jobConfig.setName(jobName);
+        Job job = jet1.newJob(p, jobConfig);
+
+        // Register JobStatusListeners
+        job.addStatusListener(event -> System.out.printf("Job status changed: %s -> %s. User requested? %b%n",
+                event.getPreviousStatus(), event.getNewStatus(), event.isUserRequested()));
+        waitForStatus(job, JobStatus.RUNNING);
+        job.addStatusListener(event -> { if (event.getNewStatus() == JobStatus.RUNNING) runCount++; });
```
burakgok, line 42: Move these lines after printing "Job 'sampleName' is submitted."

burakgok (changes requested):

burakgok (approved):

PR #584, opened 2023-07-14
Remove redundant destroy call for the same distributed object

arodionov:
Remove redundant destroy for the same object.
There is no need to call destroy() several times if they are applied to the same distributed object.

### DIFF
--- a/cp-subsystem/fenced-lock-monotonic-fencing-tokens/src/main/java/com/hazelcast/codesamples/cp/fencedlock/MonotonicFencingTokens.java
+++ b/cp-subsystem/fenced-lock-monotonic-fencing-tokens/src/main/java/com/hazelcast/codesamples/cp/fencedlock/MonotonicFencingTokens.java
@@ -53,7 +53,6 @@ public class MonotonicFencingTokens {
 
         // always destroy CP Subsystem data structures otherwise it can lead to a memory leak
         hz1Lock.destroy();
-        hz2Lock.destroy();
 
         hz1.getLifecycleService().terminate();
         hz2.getLifecycleService().terminate();
--- a/cp-subsystem/fenced-lock-using-fencing-tokens/src/main/java/com/hazelcast/codesamples/cp/fencedlock/FencingOffStaleLockHolders.java
+++ b/cp-subsystem/fenced-lock-using-fencing-tokens/src/main/java/com/hazelcast/codesamples/cp/fencedlock/FencingOffStaleLockHolders.java
@@ -132,7 +132,6 @@ public class FencingOffStaleLockHolders {
 
         // always destroy CP Subsystem data structures otherwise it can lead to a memory leak
         hz1Lock.destroy();
-        hz2Lock.destroy();
 
         hz1.getLifecycleService().terminate();
         hz2.getLifecycleService().terminate();


cp-subsystem/fenced-lock-using-fencing-tokens/src/main/java/com/hazelcast/codesamples/cp/fencedlock/FencingOffStaleLockHolders.java
```diff
@@ -132,7 +132,6 @@ public static void main(String[] args) throws InterruptedException, ExecutionExc
 
         // always destroy CP Subsystem data structures otherwise it can lead to a memory leak
         hz1Lock.destroy();
-        hz2Lock.destroy();
```
avtarraikmo, line 135: FencedLock hz1Lock = hz1.getCPSubsystem().getLock(lockName);
        FencedLock hz2Lock = hz2.getCPSubsystem().getLock(lockName);

Don't we have two to destroy here? Are you sure we need to only destroy only 1?

cp-subsystem/fenced-lock-monotonic-fencing-tokens/src/main/java/com/hazelcast/codesamples/cp/fencedlock/MonotonicFencingTokens.java
```diff
@@ -53,7 +53,6 @@ public static void main(String[] args) {
 
         // always destroy CP Subsystem data structures otherwise it can lead to a memory leak
         hz1Lock.destroy();
-        hz2Lock.destroy();
```
avtarraikmo, line 56: // The lock switches from the available state to the held state.
        FencedLock hz1Lock = hz1.getCPSubsystem().getLock("my-lock");

        // The lock switches from the available state to the held state.
        FencedLock hz2Lock = hz2.getCPSubsystem().getLock("my-lock");
        
Although we have two lock objects, the underlying reference is to the same lock in the CP Subsystem. 
Do we need to destroy both or only one?

cp-subsystem/fenced-lock-using-fencing-tokens/src/main/java/com/hazelcast/codesamples/cp/fencedlock/FencingOffStaleLockHolders.java
avtarraikmo: Although we have two lock objects, the underlying reference is to the same lock in the CP Subsystem. 
Do we need to destroy both or only one?

cp-subsystem/fenced-lock-monotonic-fencing-tokens/src/main/java/com/hazelcast/codesamples/cp/fencedlock/MonotonicFencingTokens.java
avtarraikmo: If only one - suggest we put a comment next to it, to explain why. I can imagine someone (maybe just me), reading through it quickly and not picking up on that first time.

Thoughts?

avtarraikmo (approved):
Thanks!

PR #604, opened 2023-12-18
Revert JDK 17 cleanup

orcunc:
I had to revert everything because the previous PR has also upgraded to SpringBoot 3 which requires JDK 17. 
So it was not possible to just downgrade java.version to 11


orcunc:
run-lab-run

PR #614, opened 2024-02-05
Add UCN dynamic configuration with a JAR code sample

mehmettokgoz:
Building on the previous UCN code samples, this PR introduces a new code sample that dynamically configures a namespace using a JAR.


JackPGreen (approved):

enterprise/user-code-namespaces/README.md
```diff
@@ -96,4 +96,42 @@ mvn clean compile exec:java -Dexec.mainClass=com.hazelcast.namespaces.dyamicconf
...
+Run Member:
+
+```
+cd user-code-namespaces-member
+mvn clean compile exec:java -Dexec.mainClass=com.hazelcast.namespace.dynamic.Member
+```
+
+**Member code walkthrough:**
+
+- Enables user code namespaces feature on config.
```
JamesHazelcast, line 115:
```suggestion
- Enables user code namespaces feature on `Config`.
```

enterprise/user-code-namespaces/README.md
```diff
@@ -96,4 +96,42 @@ mvn clean compile exec:java -Dexec.mainClass=com.hazelcast.namespaces.dyamicconf
...
+- Creates a new `UserCodeNamespaceConfig` named `ucn1` with no resources.
+*Note - this config must exist on the running instance before the client can add to it.*
```
JamesHazelcast, line 117: This is not correct - clients can dynamically create a new Namespace instance at runtime (provided they have the required permissions)

enterprise/user-code-namespaces/README.md
```diff
@@ -96,4 +96,42 @@ mvn clean compile exec:java -Dexec.mainClass=com.hazelcast.namespaces.dyamicconf
...
+```
+cd user-code-namespaces-client
+mvn clean compile exec:java -Dexec.mainClass=com.hazelcast.namespaces.dynamicconfig.jar.Client
+```
+
+**Client Code walkthrough:**
+
+- Starts the client.
+- Creates a new `UserCodeNamespaceConfig` with name `ucn1` and adds the external JAR URL.
+- Gets the hazelcast config from the client instance and adds the `UserCodeNamespaceConfig` to the `Config`.
```
JamesHazelcast, line 134:
```suggestion
- Gets the Hazelcast config from the client instance and adds the `UserCodeNamespaceConfig` to the `Config`.
```

enterprise/user-code-namespaces/user-code-namespaces-client/src/main/java/com/hazelcast/namespaces/dynamicconfig/jar/Client.java
```diff
@@ -0,0 +1,41 @@
...
+
+import com.hazelcast.client.HazelcastClient;
+import com.hazelcast.config.UserCodeNamespaceConfig;
+import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.map.EntryProcessor;
+import usercodenamespaces.IncrementingEntryProcessor;
+
+import java.io.File;
+import java.net.URL;
+public class Client {
```
JamesHazelcast, line 11: Nit: missing new line between imports and class declaration

enterprise/user-code-namespaces/user-code-namespaces-client/src/main/java/com/hazelcast/namespaces/dynamicconfig/jar/Client.java
```diff
@@ -0,0 +1,41 @@
...
+        client.getMap("map1").executeOnKey("key", entryProcessor);
+
+        // will print incremented value
+        System.out.println(client.getMap("map1").get("key"));
+        client.shutdown();
+    }
+
+    /**
+     * Helper code to load the example jar from outside the classpath
+     * @return the URL of the jar.
```
JamesHazelcast, line 32: Nit: the helper method exists in the other code sample because obtaining a `Class` object from outside the classpath is not trivial - however retrieving a JAR URL is imo, so not sure we need a helper method for it.

JamesHazelcast (changes requested):
Thanks for putting this together Mehmet, it's much appreciated! Just a few comments from me, small bits :)

enterprise/user-code-namespaces/README.md
mehmettokgoz: I copied this warning from `3. Dynamically configured namespace with a single class.` section.  

@gareth-johnston  wrote that note, should we also remove it?

JamesHazelcast: Yes I think so, unless @gareth-johnston I'm missing something here?

gareth-johnston: I think the point this was trying to make, admittedly not clearly enough, is the following: 

We start up a member which statically creates a map, the map references an empty namespace `ucn1` that has just been created on the member.

We run a client that wants to configure `ucn1` whilst member is running. This is where this note comes in ->  *Note - this config must exist on the running instance before the client can add to it.*

The point being that the IMap which is statically configured references a namespace - so in this case it must be pre-configured before the client can do a replace call.

Does that make sense? I can see how it is possibly not clear enough

JamesHazelcast: @mehmettokgoz if you're able to apply this change, we can get this PR merged 👍

JamesHazelcast (approved):

PR #628, opened 2024-04-08
Updated CP samples to EE [HZ-4480]

arodionov:


gbarnett-hz (approved):
I ran a random example just to make sure -- `cp-semaphore`.

Nothing to do with this PR but from what I observed: 

- run `start.sh` in 3 terminals
- cycles through some release-acq

Two members exit at the end, one member just sits there. First time I run this so not sure if this is expected or not.

arodionov:
Yes, there is a known issue with the last CP Member.

PR #636, opened 2024-05-22
Make default Hazelcast version, last released version (i.e. not `SNAPSHOT`).

JackPGreen:
[`SNAPSHOT` Hazelcast builds are no longer publicly accessible](https://hazelcast.atlassian.net/browse/DI-100), which leads to build failures as `code-samples` [relies on `SNAPSHOT`s since virtually the beginning](https://github.com/hazelcast/hazelcast-code-samples/commit/c1121d4ac9169d56ed2879c337e9c345b1216322).

It makes more sense for examples of how to use Hazelcast to be backed by production-quality dependencies, i.e. a released version.

As such, downgraded to the latest released version of Hazelcast - will be kept up-to-date via [Dependabot](https://github.com/hazelcast/hazelcast-code-samples/pull/574).

[Slack discussion](https://hazelcast.slack.com/archives/C05LM8B80UT/p1716380405098929)

### DIFF
--- a/jet/wordcount-compute-isolation/pom.xml
+++ b/jet/wordcount-compute-isolation/pom.xml
@@ -84,5 +84,7 @@
         <!-- needed for checkstyle/findbugs -->
         <main.basedir>${project.parent.parent.basedir}</main.basedir>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+		<!-- TODO Once Hazelcast v5.5 is released, can become non-SNAPSHOT -->
+		<hazelcast.version>5.5.0-SNAPSHOT</hazelcast.version>
     </properties>
 </project>
--- a/jet/wordcount-compute-isolation/pom.xml
+++ b/jet/wordcount-compute-isolation/pom.xml
@@ -84,7 +84,7 @@
         <!-- needed for checkstyle/findbugs -->
         <main.basedir>${project.parent.parent.basedir}</main.basedir>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-		<!-- TODO Once Hazelcast v5.5 is released, can become non-SNAPSHOT -->
-		<hazelcast.version>5.5.0-SNAPSHOT</hazelcast.version>
+        <!-- TODO Once Hazelcast v5.5 is released, can become non-SNAPSHOT -->
+        <hazelcast.version>5.5.0-SNAPSHOT</hazelcast.version>
     </properties>
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -16,7 +16,7 @@
         <main.basedir>${project.basedir}</main.basedir>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
-        <hazelcast.version>5.5.0-SNAPSHOT</hazelcast.version>
+        <hazelcast.version>5.4.0</hazelcast.version>
         <java.version>17</java.version>
         <hazelcast-jclouds.version>3.7.2</hazelcast-jclouds.version>
         <hazelcast.hibernate.version>5.1.0</hazelcast.hibernate.version>


jet/wordcount-compute-isolation/pom.xml
```diff
@@ -84,5 +84,7 @@
         <!-- needed for checkstyle/findbugs -->
         <main.basedir>${project.parent.parent.basedir}</main.basedir>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+		<!-- TODO Once Hazelcast v5.5 is released, can become non-SNAPSHOT -->
+		<hazelcast.version>5.5.0-SNAPSHOT</hazelcast.version>
```
nishaatr, line 88: How do we track this ?
In the meantime, does code-samples compile fully?
I presume SNAPSHOT is required for some changes as 5.4.0 doesn't have the change?

pom.xml
```diff
@@ -16,7 +16,7 @@
         <main.basedir>${project.basedir}</main.basedir>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
-        <hazelcast.version>5.5.0-SNAPSHOT</hazelcast.version>
+        <hazelcast.version>5.4.0</hazelcast.version>
```
nishaatr, line 19: note: Release Pipeline will update master when we next do MAJOR or MINOR

nishaatr (changes requested):

pom.xml
JackPGreen: > note: Release Pipeline will update master when we next do MAJOR or MINOR

Are you sure it will update? I thought it only updated the current `master` `SNAPSHOT` version to new one (i.e. `5.5.0-SNAPSHOT` -> `5.6.-SNAPSHOT`), not from previously released to newly released?

Some of the files use older versions of Hazelcast and they haven't been updated.

jet/wordcount-compute-isolation/pom.xml
JackPGreen: > How do we track this ?

I don't know, I have no mechanism for enforcing this.

> In the meantime, does code-samples compile fully?

Yes - the PR builder is passing.
This project is using `5.5.0-SNAPSHOT` of EE, which is available in the repo so it's ok.

> I presume SNAPSHOT is required for some changes as 5.4.0 doesn't have the change?

This is demonstrating new 5.5 functionality.
It's _likely_ newer functionality will be EE centric so we shouldn't really have many problems.

nishaatr: thanks

pom.xml
nishaatr: You are right and thats what I meant like this one https://github.com/hazelcast/hazelcast-code-samples/commit/3a963a683af992b325db8c41f52edef9b43d76cf. Release pipeline only updates <hazelcast.version> which should trickle down to sub modules 

So I am thinking if dependabot updates this and we make MAJOR/MINOR, will dependabot create a master PR, lets say we tag v5.6.0 in the future? I am sure it will but just want to check!

JackPGreen: > You are right and thats what I meant like this one [3a963a6](https://github.com/hazelcast/hazelcast-code-samples/commit/3a963a683af992b325db8c41f52edef9b43d76cf). Release pipeline only updates <hazelcast.version> which should trickle down to sub modules

I think it only updates it if it's the specific previous `SNAPSHOT` (or whatever version is specified when you runn it), so now here it will do nothing.
 
> So I am thinking if dependabot updates this and we make MAJOR/MINOR, will dependabot create a master PR, lets say we tag v5.6.0 in the future? I am sure it will but just want to check!

This change (I think!) makes release pipeline not affect the hazelcast version going forwards. And Dependabot will (and does) think Hazelcast is just-another-dependency and suggests updates as and and when there are any publically.

nishaatr: I think you are right but if nothing is updated, then I think commit might fail as looks like currently `allowEmpty` is not passed and set to `false`. Worth checking!

JackPGreen: > I think you are right but if nothing is updated, then I think commit might fail as looks like currently `allowEmpty` is not passed and set to `false`. Worth checking!

Good point - are you able to amend as part of the other stuff your doing, please?

nishaatr: yes sure. just remind in few days but I am sure I will remember

nishaatr (approved):

nishaatr:
FYI...

I have made some changes under https://github.com/hazelcast/hazelcast-pipeline-library/pull/249 to support `git --allow-empty` for code-samples. 

However, I just tested code-samples locally and it fails if `<hazelcast.version>` is set to a release like 5.4.0 in `master` and not SNAPSHOT. This happens when I try to create a MINOR release branch cut from master and the failed check is [here](https://github.com/hazelcast/hazelcast-pipeline-library/blob/master/src/com/hazelcast/qe/pipeline/release/maven/VersionResolve.groovy#L119)

image

I am just highlighting here as the use of SNAPSHOT (in code-samples and deployment) is still an open question. If we decide to use last release in code-samples, then I will need to fix above 

Then the questions arises whether we need this PR to auto-update `<hazelcast.version>` for code-samples? I will need to make changes to fix this in any case, and my thinking is, may be best we do this in the Release Pipeline to set required version in `release-branch`, `z-branch`  and `master` so its in one place and makes release instructions little simpler. We can discuss this further once we have a way forward on the question where we should deploy SNAPSHOT or not

olukas:
> It makes more sense for examples of how to use Hazelcast to be backed by production-quality dependencies, i.e. a released version.

I disagree. If somebody wants to use some certain version of Hazelcast for code samples they should use related tag (for example https://github.com/hazelcast/hazelcast-code-samples/blob/v5.4.0/pom.xml#L19). If somebody decides to use `master` branch of code samples (which is latest version) they should use latest `master` (i.e. SNAPSHOT) of Hazelcast. 

If we will use latest released Hazelcast version for code samples (instead of SNAPSHOT) it also means we cannot add a code samples for the new features from Hazelcast `master` branch since code samples doesn't use it.

I think that if latest SNAPSHOT is no longer publicly available then proper solution is to build Hazelcast locally (from `master`) before compiling code-samples.

JackPGreen:
> > It makes more sense for examples of how to use Hazelcast to be backed by production-quality dependencies, i.e. a released version.
> 
> I disagree. If somebody wants to use some certain version of Hazelcast for code samples they should use related tag (for example https://github.com/hazelcast/hazelcast-code-samples/blob/v5.4.0/pom.xml#L19). If somebody decides to use `master` branch of code samples (which is latest version) they should use latest `master` (i.e. SNAPSHOT) of Hazelcast.

The problem (I think) is that the default branch is `master`, so out-of-the-box, it won't build.

While you _can_ use a tag to specify your targeted version (e.g. as per the docs), I don't think that's how any references to `code-samples` have been communicated to date ([rather contrived example](https://hub.docker.com/r/hazelcast/hazelcast-kubernetes)).


> If we will use latest released Hazelcast version for code samples (instead of SNAPSHOT) it also means we cannot add a code samples for the new features from Hazelcast `master` branch since code samples doesn't use it.

We can, for EE features (as there is still a `SNAPSHOT` for those versions) - I've updated `jet/wordcount-compute-isolation` to support that. But obviously not as ideal as it was.
 
> I think that if latest SNAPSHOT is no longer publicly available then proper solution is to build Hazelcast locally (from `master`) before compiling code-samples.

That would work for the PR builder (at the expense of execution time), but it wouldn't help new users.

One of _my personal_ bugbears is when you try to evaluate a product and the samples/tutorial etc doesn't work out-of-the-box.  And the fact that I got my first ever application running with Hazelcast in minutes was very nice.

kwart (approved):

PR #674, opened 2024-11-11
Migrate Jenkins PR builder to GitHub Action [DI-336]

nishaatr:
This came about at the back of removing Code Samples from Platform releases https://hazelcast.atlassian.net/browse/DI-313

TODO (need help here)
1. Sort out `HZ_LICENSEKEY`. We have these in GH so not sure which one to use
   - HAZELCAST_ENTERPRISE_KEY (may be start with this?)
   - HAZELCAST_ENTERPRISE_KEY_V7 (this will future proof perhaps?)
4. Limit who runs this via the action (checking for org membership) or via the GitHub repo config?
5. Remove `-DskipTests` once above is sorted

Fixes: [DI-336](https://hazelcast.atlassian.net/browse/DI-336)

[DI-336]: https://hazelcast.atlassian.net/browse/DI-336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #679, opened 2024-11-12
Add PR membership check  [DI-336]

nishaatr:
Follow-up from https://github.com/hazelcast/hazelcast-code-samples/pull/674

I tested this:

INTERNAL user: https://github.com/hazelcast/pr-test/pull/1 (passes)
image

EXTERNAL user: https://github.com/hazelcast/pr-test/pull/2 (fails)

image
image

After merge, enable Actions again https://github.com/hazelcast/hazelcast-code-samples/settings/actions

NOTE: will need to merge without PR builder as its disabled. I will retest again after merge

### DIFF
--- a/.github/workflows/builder.yaml
+++ b/.github/workflows/builder.yaml
@@ -19,7 +19,7 @@ jobs:
           token: ${{ secrets.GH_TOKEN }}
   pr-builder:
     runs-on: ubuntu-latest
-    needs: [check_for_membership]
+    needs: check_for_membership
     steps:
       - name: Detect untrusted community PR
         if: ${{ needs.check_for_membership.outputs.check-result == 'false' }}
--- a/.github/workflows/builder.yaml
+++ b/.github/workflows/builder.yaml
@@ -36,4 +36,10 @@ jobs:
       - name: Build and test
         run: |
           ${RUNNER_DEBUG:+set -x}
-          ls -al
+          mvn \
+            --batch-mode \
+            --errors \
+            --no-transfer-progress \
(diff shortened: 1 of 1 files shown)


.github/workflows/builder.yaml
```diff
@@ -4,9 +4,28 @@ on:
...
+    name: Check PR author membership
+    outputs:
+      check-result: ${{ steps.composite.outputs.check-result }}
+    steps: 
+      - name: Action for membership check
+        id: composite
+        uses: hazelcast/hazelcast-tpm/membership@main
+        with:
+          organization-name: 'hazelcast'
+          member-name: ${{ github.event.pull_request.head.repo.owner.login }}
```
JackPGreen, line 18:
```suggestion
          member-name: ${{ github.actor }}
```

Otherwise I _think_ this will always fail.

.github/workflows/builder.yaml
```diff
@@ -4,9 +4,28 @@ on:
...
+          token: ${{ secrets.GH_TOKEN }}
   pr-builder:
     runs-on: ubuntu-latest
+    needs: [check_for_membership]
```
JackPGreen, line 22:
```suggestion
    needs: check_for_membership
```

.github/workflows/builder.yaml
nishaatr: Good catch. I was wondering about this
Tested by me making a change in the test user branch and builder passes. For sanity I made subsequent chaneg with the test user and it fails again as expect
note: does not work when `Re-run all jobs`. Both the above values point to the test user

JackPGreen (approved):

PR #694, opened 2024-12-01
Hazelcast integration with Hikari connection pool

rajangare:
Hazelcast Integration with Hikari Connection Pool
To create Hikari connection pool, first create database table mapping with Hazelcast in-memory map.
In this demo, I have used EMPLOYEE table created on H2 DB, which is mapped with employee-map.

- Create Hikari data source from HikariConfig, with database credentials and connection pool configuration e.g. pool size, connection timeout etc.
- Get the connection from Hikari Data Source.
- Implement the Map Store Factory from MapLoader<K, V>, MapStore<K, V> and override all the necessary methods.
- Register the MapStore into hazelcast.xml config.
   
  
    <hazelcast>
     ...
        <map name="empaloyee_map">
           <map-store enabled="true">
             <class-name>com.demo.maps.EmployeeMapStore</class-name>
           </map-store>
        </map>
     ...
    </hazelcast>

Database connection information can be externalized from hazelcast.xml. Build the above classes in a jar and copy them inside Hazelcast user-lib directory under the Hazelcast installation.


frant-hartm (changes requested):

yuce (changes requested):
Thanks for the PR!
Just a few ideas...

yuce:
@rajangare Latest changes don't make this class thread-safe, but it would be good if you could add a comment that mentions this issue.

rajangare:
> @rajangare Latest changes don't make this class thread-safe, but it would be good if you could add a comment that mentions this issue.

@yuce added the comments on top, for thread-safe. do you have any suggestion to make it thread-safe please help here.

PR #700, opened 2024-12-20
Fix HazelcastSpringClient

TomaszGaweda:
1. Removed licensed CP stuff from applicationContext.xml
2. Updated XSD version
3. Added dependency to hazelcast-spring as it was some problem with finding 5.5 XSD


Patras3 (approved):
LGTM, thanks!

PR #703, opened 2025-02-09
Sample code for standalone java client [PA-627]

roman-morenko:


roman-morenko:
- Created new maven module client-standalone as child module for clients module
- Created Client class. In main method we create instance of standalone java client, then create map, in loop we put values to map then get value by key and print it.
- Created scripts for starting client and server for Linux/MacOS and Windows 
- Created REAME.ME with instructions how to run sample.

PR #720, opened 2025-06-13
DOC-500 add Jet payments demo

Rob-Hazelcast:
Import https://github.com/knurl/jetpayments to official repo.

https://hazelcast.atlassian.net/browse/DOC-500

Checklist:
- [x] Request reviewers if possible
- [x] README.md or README.adoc is created with a description _what_ example does and _how_ to run it
- [ ] Add tests that reproduce the associated tutorial or similar scenario


jet-payments-kotlin/LICENSE
JackPGreen, line 1: Do we need / should we have this?

jet-payments-kotlin/build.gradle.kts
JackPGreen, line 1: This won't be picked up when the `hazelcast-code-samples` project is built

I think we'd need to:
- convert this (sub) project to build using Maven
- add it as a module to the parent.
- remove the Gradle wrapper

...or update the build scripts to build this project in a hardcoded way. But that's bad.

oliverhowell:
HI @JackPGreen this is code that was made by a former SA and connects to a video demo explaining how to use it. We've not changed the code at all and I'm reluctant to do so not knowing the instructions in the video, hence why we didn't want to change the folder location either. With that context, do you still feel need to make these changes to upload to code-samples?

